PR #6, opened 2020-03-17
Implement payment using nanopay for reverse tuna

bufrr:


bin/exit/main.go
```diff
@@ -63,8 +63,9 @@ func main() {
 	}
 
 	if config.Reverse {
-		for serviceName := range config.Services {
-			e := tuna.NewTunaExit(config, services, wallet)
+		for serviceName, serviceInfo := range config.Services {
+			reverseEntryToExitMaxPrice, reverseExitToEntryMaxPrice, err := tuna.ParsePrice(serviceInfo.Price)
```
yilunzhang: `ReverseMaxPrice` should be used for reverse price

entry.go
```diff
@@ -11,6 +13,9 @@ import (
 	"unsafe"
 
 	nkn "github.com/nknorg/nkn-sdk-go"
+
+	"github.com/nknorg/nkn/transaction"
```
yilunzhang: Did you use gofmt or goimports to format code?

entry.go
```diff
@@ -122,46 +128,8 @@ func (te *TunaEntry) Start() {
...
-				}
-				n, err := stream.Write(txData)
-				if n == len(txData) && err == nil {
-					te.bytesInPaid = bytesIn
-					te.bytesOutPaid = bytesOut
-				}
-				stream.Close()
-			}
-		}()
+		session, err := te.getSession(false)
```
yilunzhang: It's better to call `getSession` when use it (like the code you deleted) because previous session might be closed. Also it's better to change `sendNanoPayment()` to `te. sendNanoPayment()` so you don't need to pass so many arguments here

entry.go
```diff
@@ -84,8 +89,9 @@ func (te *TunaEntry) Start() {
 			time.Sleep(1 * time.Second)
 			continue
 		}
-
-		tcpPorts, err := te.listenTCP(te.ListenIP, te.Service.TCP)
+		reverseBytesIn := make([]uint64, 256)
+		reverseBytesOut := make([]uint64, 256)
+		tcpPorts, err := te.listenTCP(te.ListenIP, te.Service.TCP, reverseBytesIn, reverseBytesOut)
```
yilunzhang: We can make `reverseBytesIn` and `reverseBytesOut` properties of `TuneEntry` just like `bytesIn`, `bytesInPaid`, `bytesOut`, `bytesOutPaid`

entry.go
bufrr: So we need to add `getSession` function to the `TunaExit` also?

yilunzhang: There is already one, see https://github.com/nknorg/tuna/pull/6/files#diff-32f22e4789509e4288e8ee93f532fbd0L149

entry.go
```diff
@@ -208,8 +179,68 @@ func (te *TunaEntry) StartReverse(stream *smux.Stream) error {
...
+
+		lastComputed := common.Fixed64(0)
+		lastClaimed := common.Fixed64(0)
+		lastUpdate := time.Now()
+		claimInterval := time.Duration(te.Config.ReverseClaimInterval) * time.Second
+		//fmt.Println("claimInterval:", time.Duration(te.Config.ReverseClaimInterval))
+		onErr := nkn.NewOnError(1, nil)
+		isClosed := false
+
+		npc, err := te.Wallet.NewNanoPayClaimer(int32(10000), onErr, te.Config.ReverseBeneficiaryAddr)
```
yilunzhang: Should remove the hard coded constant

entry.go
```diff
@@ -208,8 +179,68 @@ func (te *TunaEntry) StartReverse(stream *smux.Stream) error {
...
+		if err != nil {
+			log.Println(err)
```
yilunzhang: We should return here or use `log.Fatal`

entry.go
```diff
@@ -223,6 +254,53 @@ func (te *TunaEntry) StartReverse(stream *smux.Stream) error {
 	return nil
 }
 
+func (te *TunaEntry) nanoPayClaim(stream *smux.Stream, npc *nkn.NanoPayClaimer, reverseBytesIn, reverseBytesOut []uint64, lastComputed, lastClaimed common.Fixed64, lastUpdate time.Time) {
```
yilunzhang: lastComputed, lastClaimed, lastUpdate should be pointer

entry.go
yilunzhang: Nevermind, just realized you are trying to combine duplicate code

yilunzhang: You can use something like in entry:
```
go func() {
for {
    session, err := te.getSession(false)
    sendNanoPayment(session, ...)
}
}()
```
basically extract the for loop outside the function

entry.go
yilunzhang: Argument name should not contain reverse as it's used for both forward and reverse mode

entry.go
```diff
@@ -223,6 +254,53 @@ func (te *TunaEntry) StartReverse(stream *smux.Stream) error {
...
+		in := atomic.LoadUint64(&reverseBytesIn[i])
+		out := atomic.LoadUint64(&reverseBytesOut[i])
+		if in == 0 && out == 0 {
+			continue
+		}
+		//service, err := te.getService(byte(i))
+		if err != nil {
+			continue
+		}
+		entryToExitPrice, exitToEntryPrice := common.Fixed64(100000), common.Fixed64(100000)
```
yilunzhang: Parse price

entry.go
```diff
@@ -300,9 +378,13 @@ func (te *TunaEntry) listenTCP(ip net.IP, ports []int) ([]int, error) {
 					time.Sleep(time.Second)
 					continue
 				}
-
-				go Pipe(stream, conn, &te.bytesOut)
-				go Pipe(conn, stream, &te.bytesIn)
+				if te.Config.Reverse {
+					go Pipe(stream, conn, &reverseBytesIn[_port])
```
yilunzhang: Why is `reverseBytesIn` an array?

entry.go
```diff
@@ -223,6 +254,53 @@ func (te *TunaEntry) StartReverse(stream *smux.Stream) error {
...
+		}
+		//service, err := te.getService(byte(i))
+		if err != nil {
+			continue
+		}
+		entryToExitPrice, exitToEntryPrice := common.Fixed64(100000), common.Fixed64(100000)
+		if err != nil {
+			continue
+		}
+		totalCost += entryToExitPrice*common.Fixed64(in)/TrafficUnit + exitToEntryPrice*common.Fixed64(out)/TrafficUnit
```
yilunzhang: For forward, in means entry to exit
For exit, in means exit to entry
So we should use `entryToExitBytes` and `exitToEntryBytes` instead of `bytesIn` and `bytesOut` in this function

entry.go
yilunzhang: And you forget to handle error here

entry.go
bufrr: To calculate mutiple exit client's traffic statistics

yilunzhang: One TuneEntry is only for one client (exit)

bin/exit/main.go
```diff
@@ -63,8 +63,9 @@ func main() {
 	}
 
 	if config.Reverse {
-		for serviceName := range config.Services {
-			e := tuna.NewTunaExit(config, services, wallet)
+		for serviceName, serviceInfo := range config.Services {
+			reverseEntryToExitPrice, reverseExitToEntryPrice, err := tuna.ParsePrice(serviceInfo.Price)
```
yilunzhang: This is still not fixed. You should parse `ReverseMaxPrice` instead of `serviceInfo.Price`. `ReverseMaxPrice` is the highest price exit will accept, `serviceInfo.Price` is the price exit will charge when exit is service provider (not reverse mode).

entry.go
```diff
@@ -41,7 +44,7 @@ type EntryConfiguration struct {
 
 type TunaEntry struct {
 	*Common
-	config             *EntryConfiguration
+	Config             *EntryConfiguration
```
yilunzhang: Why we change this config to public variable?

entry.go
```diff
@@ -108,16 +113,13 @@ func (te *TunaEntry) Start() {
 			if err != nil {
 				return
 			}
-			stream, err := session.OpenStream()
-			if err != nil {
-				return
-			}
-			stream.Close()
+
```
yilunzhang: This is still not fixed: you should NOT remove this part. This is needed to keep TCP open if no traffic is going through tuna in the forward mode.

entry.go
```diff
@@ -108,16 +113,13 @@ func (te *TunaEntry) Start() {
...
 			for {
 				_, err = session.AcceptStream()
 				if err != nil {
 					log.Println("Close connection:", err)
-					return
+					time.Sleep(time.Second * 1)
```
yilunzhang: When a TCP accept returns error, you should return instead of continue because it means the connection is closed or cannot accept connection anymore.

entry.go
```diff
@@ -126,43 +128,14 @@ func (te *TunaEntry) Start() {
...
-				if err != nil {
-					continue
-				}
-				n, err := stream.Write(txData)
-				if n == len(txData) && err == nil {
-					te.bytesInPaid = bytesIn
-					te.bytesOutPaid = bytesOut
-				}
-				stream.Close()
+				go sendNanoPayment(np, session, te.Wallet, &te.bytesIn, &te.bytesOut, &te.bytesInPaid, &te.bytesOutPaid, te.Common, te.Config.NanoPayFee)
```
yilunzhang: no need to use goroutine here

entry.go
```diff
@@ -208,13 +181,67 @@ func (te *TunaEntry) StartReverse(stream *smux.Stream) error {
...
+		lastClaimed := common.Fixed64(0)
+		lastUpdate := time.Now()
+		claimInterval := time.Duration(te.Config.ReverseClaimInterval) * time.Second
+		onErr := nkn.NewOnError(1, nil)
+		isClosed := false
+
+		npc, err := te.Wallet.NewNanoPayClaimer(int32(claimInterval/time.Millisecond), onErr, te.Config.ReverseBeneficiaryAddr)
+		if err != nil {
+			log.Println(err)
+			return
```
yilunzhang: We need to call `te.close()` before return because we can no longer accept payment for this tuna channel.

entry.go
```diff
@@ -208,13 +181,67 @@ func (te *TunaEntry) StartReverse(stream *smux.Stream) error {
...
+						break
+					}
+					continue
+				}
+			}
+		}()
+
+		go func() {
+			for {
+				time.Sleep(time.Second * 10)
```
yilunzhang: Hard coded constant

entry.go
```diff
@@ -208,13 +181,67 @@ func (te *TunaEntry) StartReverse(stream *smux.Stream) error {
...
+			return
+		}
+
+		go func() {
+			for {
+				err := <-onErr.C
+				if err != nil {
+					log.Println("Couldn't claim nano pay:", err)
+					if npc.IsClosed() {
+						break
```
yilunzhang: You shouldn't break without close session and set isClosed to true because this npc will never be able to claim any nanopay updates, which means entry will not get any rewards

entry.go
```diff
@@ -208,13 +181,67 @@ func (te *TunaEntry) StartReverse(stream *smux.Stream) error {
 			return
 		}
 		stream.Close()
+
+		lastComputed := common.Fixed64(0)
```
yilunzhang: L185-L233 is a large chunk of duplicate code in exit `handleSession` function. You need to find a way to combine theses codes together.

tuna.go
```diff
@@ -560,3 +562,44 @@ func LoadOrCreateAccount(walletFile, passwordFile string) (*vault.Account, error
 	}
 	return wallet.GetDefaultAccount()
 }
+
+func sendNanoPayment(np *nkn.NanoPay, session *smux.Session, w *nkn.Wallet, inBytes, outBytes, inBytesPaid, outBytesPaid *uint64, c *Common, nanoPayFee string) {
```
yilunzhang: `sendNanoPayment` is a function shared by entry and exit. As I said earlier, `in` and `out` have different meaning on entry and exit side: at entry, `in` means exit to entry, while at exit, `in` means entry to exit. So we need to change all `in` and `out` related variables in this function to `entryToExitXXX` and `exitToEntryXXX`.

exit.go
```diff
@@ -545,6 +554,14 @@ func (te *TunaExit) StartReverse(serviceName string) error {
 				te.readUDP()
 			}
 
+			go func() {
+				var np *nkn.NanoPay
+				for {
+					time.Sleep(DefaultNanoPayUpdateInterval)
+					go sendNanoPayment(np, session, te.Wallet, &te.reverseBytesIn, &te.reverseBytesOut, &te.reverseBytesInPaid, &te.reverseBytesOutPaid, te.Common, "0")
```
yilunzhang: 1. No need to use goroutine here.
2. It's better for `sendNanoPayment` function to return error.
3. It's better to check if `te.closeChan` is closed in the beginning of each iteration, and return if it's closed to avoid goroutine leak.

tuna.go
yilunzhang: Also it's better for this function to return error, so the caller can decide what to do

exit.go
```diff
@@ -39,27 +39,40 @@ type ExitConfiguration struct {
...
-	common           *Common
-	tcpListener      net.Listener
-	udpConn          *net.UDPConn
-	reverseIp        net.IP
-	reverseTcp       []int
-	reverseUdp       []int
-	onEntryConnected func()
-	closeChan        chan struct{}
+	config              *ExitConfiguration
+	Wallet              *nkn.Wallet
```
yilunzhang: It's better to remove wallet and use wallet in common to be consistent with entry

exit.go
```diff
@@ -39,27 +39,40 @@ type ExitConfiguration struct {
...
+	services            []Service
+	serviceConn         *cache.Cache
+	Common              *Common
```
yilunzhang: It's better to keep Common definition the same in entry and exit.

exit.go
```diff
@@ -39,27 +39,40 @@ type ExitConfiguration struct {
...
+	reverseBytesInPaid  uint64
+	reverseBytesOutPaid uint64
 }
 
-func NewTunaExit(config *ExitConfiguration, services []Service, wallet *nkn.Wallet) *TunaExit {
+func NewTunaExit(config *ExitConfiguration, services []Service, reverseEntryToExitPrice, reverseExitToEntryPrice common.Fixed64, wallet *nkn.Wallet) *TunaExit {
 	return &TunaExit{
+		Common: &Common{
+			entryToExitPrice:   reverseEntryToExitPrice,
+			exitToEntryPrice:   reverseExitToEntryPrice,
```
yilunzhang: This should be `EntryToExitMaxPrice` and `ExitToEntryMaxPrice` instead of `entryToExitPrice` and `exitToEntryPrice`.
Note that max price and price are different:
Max price is user's max acceptable price (upper limit).
price is provider's asking price (actual price).

exit.go
```diff
@@ -39,27 +39,40 @@ type ExitConfiguration struct {
...
+	onEntryConnected    func()
+	closeChan           chan struct{}
+	reverseBytesIn      uint64
+	reverseBytesOut     uint64
+	reverseBytesInPaid  uint64
+	reverseBytesOutPaid uint64
 }
 
-func NewTunaExit(config *ExitConfiguration, services []Service, wallet *nkn.Wallet) *TunaExit {
+func NewTunaExit(config *ExitConfiguration, services []Service, reverseEntryToExitPrice, reverseExitToEntryPrice common.Fixed64, wallet *nkn.Wallet) *TunaExit {
```
yilunzhang: should be `reverseEntryToExitMaxPrice` and `reverseExitToEntryMaxPrice`

tuna.go
```diff
@@ -560,3 +562,44 @@ func LoadOrCreateAccount(walletFile, passwordFile string) (*vault.Account, error
...
+
+	bytesIn := atomic.LoadUint64(inBytes)
+	bytesOut := atomic.LoadUint64(outBytes)
+
+	delta := c.exitToEntryPrice*common.Fixed64(bytesIn-*inBytesPaid)/TrafficUnit + c.entryToExitPrice*common.Fixed64(bytesOut-*outBytesPaid)/TrafficUnit
```
yilunzhang: You should use `common.GetPrice()` instead of reading `c.entryToExitPrice` and `c.entryToExitPrice` directly because there is no lock here

exit.go
```diff
@@ -176,16 +189,8 @@ func (te *TunaExit) handleSession(session *smux.Session, conn net.Conn) {
...
-						continue
-					}
-					serviceInfo := te.config.Services[service.Name]
-					entryToExitPrice, exitToEntryPrice, err := ParsePrice(serviceInfo.Price)
-					if err != nil {
-						continue
-					}
-					totalCost += entryToExitPrice*common.Fixed64(in)/TrafficUnit + exitToEntryPrice*common.Fixed64(out)/TrafficUnit
+
+					totalCost += te.Common.entryToExitPrice*common.Fixed64(in)/TrafficUnit + te.Common.exitToEntryPrice*common.Fixed64(out)/TrafficUnit
```
yilunzhang: This is not correct. On the exit side, each service can charge for different price. `te.Common.entryToExitPrice` should only be used on user side, not service provider side.

entry.go
```diff
@@ -223,6 +250,44 @@ func (te *TunaEntry) StartReverse(stream *smux.Stream) error {
 	return nil
 }
 
+func (te *TunaEntry) nanoPayClaim(stream *smux.Stream, npc *nkn.NanoPayClaimer, lastComputed, lastClaimed *common.Fixed64, lastUpdate *time.Time) {
```
yilunzhang: This function is still a large chunk of duplicate code compared to exit. It needs to be combined.

entry.go
```diff
@@ -223,6 +250,44 @@ func (te *TunaEntry) StartReverse(stream *smux.Stream) error {
...
+
+	in := atomic.LoadUint64(&te.reverseBytesIn)
+	out := atomic.LoadUint64(&te.reverseBytesOut)
+	if in == 0 && out == 0 {
+		return
+	}
+
+	totalCost += te.Common.entryToExitPrice*common.Fixed64(out)/TrafficUnit + te.Common.exitToEntryPrice*common.Fixed64(in)/TrafficUnit
+
+	lastComputed = &totalCost
```
yilunzhang: This only change the pointer itself, it does not change the value at all.

entry.go
```diff
@@ -126,43 +128,33 @@ func (te *TunaEntry) Start() {
...
-				paymentReceiver := te.GetPaymentReceiver()
-				if np == nil || np.Address() != paymentReceiver {
-					var err error
-					np, err = te.Wallet.NewNanoPay(paymentReceiver, te.config.NanoPayFee, DefaultNanoPayDuration)
-					if err != nil {
-						continue
-					}
-				}
-				tx, err := np.IncrementAmount(delta.String())
+				session, err := te.getSession(false)
```
yilunzhang: I think it's better to get session after computing delta (like the original code) so we don't need to get or create session if there is no use.

entry.go
```diff
@@ -126,43 +128,33 @@ func (te *TunaEntry) Start() {
...
-				if err != nil {
+				bytesIn := atomic.LoadUint64(&te.bytesIn)
+				bytesOut := atomic.LoadUint64(&te.bytesOut)
+				entryToExitPrice, exitToEntryPrice := te.GetPrice()
+				cost := exitToEntryPrice*common.Fixed64(bytesIn-te.bytesInPaid)/TrafficUnit + entryToExitPrice*common.Fixed64(bytesOut-te.bytesOutPaid)/TrafficUnit
+				if cost == 0 {
 					continue
 				}
-				stream, err := session.OpenStream()
+				length, err := sendNanoPayment(np, session, te.Wallet, &cost, te.Common, te.config.NanoPayFee)
```
yilunzhang: Suggest to rename NanoPayment to NanoPay. NanoPay is the official name, not NanoPayment.

entry.go
```diff
@@ -126,43 +128,33 @@ func (te *TunaEntry) Start() {
...
+					return
 				}
-				n, err := stream.Write(txData)
-				if n == len(txData) && err == nil {
+				if length > 0 {
 					te.bytesInPaid = bytesIn
 					te.bytesOutPaid = bytesOut
 				}
-				stream.Close()
+				_, ok := <-te.closeChan
```
yilunzhang: You should use
```
select {
case <-te.closeChan:
    return
default:
}
```
otherwise it's going to block.
Also you should put this at the beginning of for loop so we can check tuna close even if `cost == 0`

entry.go
```diff
@@ -126,43 +128,33 @@ func (te *TunaEntry) Start() {
...
+				cost := exitToEntryPrice*common.Fixed64(bytesIn-te.bytesInPaid)/TrafficUnit + entryToExitPrice*common.Fixed64(bytesOut-te.bytesOutPaid)/TrafficUnit
+				if cost == 0 {
 					continue
 				}
-				stream, err := session.OpenStream()
+				length, err := sendNanoPayment(np, session, te.Wallet, &cost, te.Common, te.config.NanoPayFee)
 				if err != nil {
-					continue
+					log.Printf("send nano payment err: %v", err)
+					return
```
yilunzhang: We need to change this to `continue`. This goroutine should not stop unless tuna is closed

exit.go
```diff
@@ -90,51 +98,14 @@ func (te *TunaExit) handleSession(session *smux.Session, conn net.Conn) {
...
-					if npc.IsClosed() {
-						Close(session)
-						Close(conn)
-						isClosed = true
-						break
-					}
-				}
-			}
-		}()
+		go claimCheck(session, npc, onErr, &isClosed)
```
yilunzhang: conn should also be closed when claimCheck returns

exit.go
```diff
@@ -90,51 +98,14 @@ func (te *TunaExit) handleSession(session *smux.Session, conn net.Conn) {
...
-				if common.Fixed64(float64(lastComputed)*0.9) > lastClaimed {
-					log.Println("Nano pay amount covers less than 90% of total cost")
-					Close(session)
-					Close(conn)
-					isClosed = true
-					break
-				}
-			}
-		}()
+		go paymentCheck(session, claimInterval, &lastComputed, &lastClaimed, &lastUpdate, &isClosed)
```
yilunzhang: conn is not closed.

yilunzhang: One way to do this is to pass a onclose function

entry.go
```diff
@@ -208,13 +200,47 @@ func (te *TunaEntry) StartReverse(stream *smux.Stream) error {
...
+			stream, err = session.AcceptStream()
 			if err != nil {
 				log.Println("Close connection:", err)
 				te.close()
 				return
 			}
+			if len(stream.Metadata()) == 0 {
+				in := atomic.LoadUint64(&te.reverseBytesIn)
+				out := atomic.LoadUint64(&te.reverseBytesOut)
+				if in == 0 && out == 0 {
```
yilunzhang: You should call continue if price diff is zero

tuna.go
```diff
@@ -560,3 +563,92 @@ func LoadOrCreateAccount(walletFile, passwordFile string) (*vault.Account, error
...
+	}
+
+	amount, err := npc.Claim(tx)
+	if err != nil {
+		return fmt.Errorf("couldn't accept nano pay update: %v", err)
+	}
+
+	*lastComputed = *totalCost
+	lc := amount.ToFixed64()
+	lastClaimed = &lc
```
yilunzhang: Not updated

tuna.go
```diff
@@ -560,3 +563,92 @@ func LoadOrCreateAccount(walletFile, passwordFile string) (*vault.Account, error
...
+	lu := time.Now()
+	lastUpdate = &lu
```
yilunzhang: Not updated

entry.go
```diff
@@ -208,13 +200,47 @@ func (te *TunaEntry) StartReverse(stream *smux.Stream) error {
...
+					continue
+				}
+				totalCost += te.entryToExitPrice*common.Fixed64(out)/TrafficUnit + te.exitToEntryPrice*common.Fixed64(in)/TrafficUnit
+				err = nanoPayClaim(stream, npc, &lastComputed, &lastClaimed, &totalCost, &lastUpdate)
+				if err != nil {
+					log.Println("nanoPayClaim failed:", err)
+					te.close()
```
yilunzhang: No need to close on single claim failure. claimCheck and paymentCheck will close te if not updating payment for a while

exit.go
```diff
@@ -149,49 +120,26 @@ func (te *TunaExit) handleSession(session *smux.Session, conn net.Conn) {
...
+					continue
 				}
+				totalCost += entryToExitPrice*common.Fixed64(in)/TrafficUnit + exitToEntryPrice*common.Fixed64(out)/TrafficUnit
+			}
 
-				lastComputed = totalCost
-				lastClaimed = amount.ToFixed64()
-				lastUpdate = time.Now()
-			}(stream)
+			err = nanoPayClaim(stream, npc, &lastComputed, &lastClaimed, &totalCost, &lastUpdate)
```
yilunzhang: Error not handled or print

exit.go
```diff
@@ -149,49 +120,26 @@ func (te *TunaExit) handleSession(session *smux.Session, conn net.Conn) {
...
-					return
-				}
-				if len(txData) == 0 {
-					return
-				}
-				tx := new(transaction.Transaction)
-				if err := tx.Unmarshal(txData); err != nil {
-					log.Println("Couldn't unmarshal payment stream data:", err)
-					return
+			totalCost := common.Fixed64(0)
```
yilunzhang: nanoPayClaim might be blocking because it has RPC call. It's better to put this part in goroutine like before.

exit.go
```diff
@@ -264,7 +215,8 @@ func (te *TunaExit) listenTCP(port int) error {
 				continue
 			}
 
-			go te.handleSession(session, conn)
+			go te.handleSession(session)
+			Close(conn)
```
yilunzhang: Close will be called immediately

tuna.go
```diff
@@ -560,3 +563,92 @@ func LoadOrCreateAccount(walletFile, passwordFile string) (*vault.Account, error
...
+		return 0, err
+	}
+	txData := tx.ToArray()
+	stream, err := session.OpenStream()
+	if err != nil {
+		return 0, err
+	}
+	n, err := stream.Write(txData)
+	stream.Close()
+	if !(n == len(txData) && err == nil) {
```
yilunzhang: If `n != len(txData)`, it's better to return an error explicitly, so `sendNanoPayment` only needs to return an error

exit.go
```diff
@@ -545,8 +497,38 @@ func (te *TunaExit) StartReverse(serviceName string) error {
...
+					length, err := sendNanoPayment(np, session, te.Wallet, &cost, te.Common, "0")
+					if length > 0 {
+						te.reverseBytesInPaid = bytesIn
+						te.reverseBytesOutPaid = bytesOut
+					}
+					if err != nil {
+						log.Printf("send nano payment err: %v", err)
+						return
+					}
+					_, ok := <-te.closeChan
```
yilunzhang: Same as above:
1. make it non-blocking
2. put it to the beginning of for loop

exit.go
```diff
@@ -545,8 +497,38 @@ func (te *TunaExit) StartReverse(serviceName string) error {
...
+					}
+					log.Println("incurred costs：", cost)
+					length, err := sendNanoPayment(np, session, te.Wallet, &cost, te.Common, "0")
+					if length > 0 {
+						te.reverseBytesInPaid = bytesIn
+						te.reverseBytesOutPaid = bytesOut
+					}
+					if err != nil {
+						log.Printf("send nano payment err: %v", err)
+						return
```
yilunzhang: Same as above, we should use `continue` instead of `return`

exit.go
```diff
@@ -215,7 +220,7 @@ func (te *TunaExit) listenTCP(port int) error {
 				continue
 			}
 
-			go te.handleSession(session)
+			te.handleSession(session)
```
yilunzhang: Should be
```
go func() {
    te.handleSession(session)
    Close(conn)
}
```

PR #7, opened 2020-04-02
Add geo filter in config file for selecting service nodes

bufrr:


yilunzhang:
The current interface is not extendable. For example, we might want to have nearest geo location, or ip whitelist, or ip blacklist, etc. It's better to have a better design so we can add all different filters without adding function argument each time.

Also another thing: currently the user side randomly get 1 subscriber and check if it satisfies all conditions. This does not work well if most of the subscribers can not pass the filter. For example, if only 10% of the subscribers can pass the filter, we would need to make around 10 RPC calls. A better way is to get X subscribers each time and filter them locally.

util.go
```diff
@@ -1,11 +1,38 @@
...
+type Location struct {
+	Code      string  `json:"Code"`
+	Country   string  `json:"Country"`
+	City      string  `json:"City"`
+	Latitude  float32 `json:"Latitude"`
+	Longitude float32 `json:"Longitude"`
+}
+
+func (l *Location) Empty() bool {
+	return *l == Location{}
```
yilunzhang, line 23: A tiny improvement: define a variable `var emptyLocation = Location{}`, and then compare `*l == emptyLocation` to avoid unnecessary memory allocation for each `Empty()` call.

util.go
```diff
@@ -23,3 +50,33 @@ func ParsePrice(priceStr string) (common.Fixed64, common.Fixed64, error) {
 	}
 	return entryToExitPrice, exitToEntryPrice, nil
 }
+
+func getLocationFromIP2C(ip string) (*Location, error) {
+	queryUrl := IP2CUrl + ip
+	resp, err := http.Get(queryUrl)
```
yilunzhang, line 56: We should always add a timeout when making request, otherwise this line may block forever. Example: https://github.com/nknorg/nkn/blob/master/api/httpjson/client/client.go#L31

util.go
```diff
@@ -23,3 +50,33 @@ func ParsePrice(priceStr string) (common.Fixed64, common.Fixed64, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	defer resp.Body.Close()
+
+	body, err := ioutil.ReadAll(resp.Body)
```
yilunzhang, line 63: err is not handled

util.go
```diff
@@ -23,3 +50,33 @@ func ParsePrice(priceStr string) (common.Fixed64, common.Fixed64, error) {
...
+	return loc, nil
+}
+
+func parseIP2C(body string) (*Location, error) {
+	res := strings.Split(body, ";")
+	if res[0] == "0" {
+		return nil, errors.New("get ip2c result err")
+	} else if res[0] == "1" {
+		l := new(Location)
+		l.Code = res[1]
```
yilunzhang, line 77: It's better to make sure `len(res) > 3` to avoid panic

yilunzhang:
Suggest to move geo related code from `util.go` to a new file like `geo.go`

exit.go
```diff
@@ -37,6 +37,7 @@ type ExitConfiguration struct {
 	ClaimInterval        uint32                     `json:"ClaimInterval"`
 	Services             map[string]ExitServiceInfo `json:"Services"`
 	ReverseNanoPayFee    string                     `json:"ReverseNanopayfee"`
+	IPFilter             IPFilter                   `json:"IPFilter"`
```
yilunzhang, line 40: This should be `ReverseIPFilter`

entry.go
```diff
@@ -37,6 +37,7 @@ type EntryConfiguration struct {
 	ReverseSubscriptionPrefix   string                      `json:"ReverseSubscriptionPrefix"`
 	ReverseSubscriptionDuration uint32                      `json:"ReverseSubscriptionDuration"`
 	ReverseSubscriptionFee      string                      `json:"ReverseSubscriptionFee"`
+	IPFilter                    IPFilter                    `json:"IPFilter"`
```
yilunzhang, line 40: I think we need per-service IP filter. For example, as an entry, I may use 2 services: one for httpproxy (which I need an US IP address), and one for moonlight game streaming (which I need a China IP address for better latency). A single global IP filter is not enough.

tuna.go
```diff
@@ -298,7 +298,10 @@ func (c *Common) CreateServerConn(force bool) error {
 			if subscribersCount == 0 {
 				return errors.New("there is no service providers for " + c.Service.Name)
 			}
-			offset := rand.Intn(int(subscribersCount))
+			offset := rand.Intn(subscribersCount)
+			if filter != nil && !filter.Empty() {
```
yilunzhang, line 302: You can change the `Empty()` function so it can be called even when filter is nil:
```
function (filter *IPFIlter) Empty() bool {
  if filter == nil {
    return true
  }
  ...
}
```

tuna.go
```diff
@@ -311,6 +314,33 @@ func (c *Common) CreateServerConn(force bool) error {
 
 				metadata := c.GetMetadata()
 
+				if filter != nil && !filter.Empty() {
```
yilunzhang, line 317: You can change the `Empty()` function so it can be called even when filter is nil:
```
function (filter *IPFIlter) Empty() bool {
  if filter == nil {
    return true
  }
  ...
}
```

tuna.go
```diff
@@ -298,7 +298,10 @@ func (c *Common) CreateServerConn(force bool) error {
 			if subscribersCount == 0 {
 				return errors.New("there is no service providers for " + c.Service.Name)
 			}
-			offset := rand.Intn(int(subscribersCount))
+			offset := rand.Intn(subscribersCount)
+			if filter != nil && !filter.Empty() {
+				offset = 0
```
yilunzhang, line 303: This is incorrect: you will always get the first subscriber if filter is not empty.

yilunzhang: As I commented before, we should no longer getting one subscriber at a time because it's very inefficient. So L294-L308 should be changed:
1. Define a global constant, e.g. `getSubscribersBatchSize` to some value like 32
2. `offset := rand.Intn(int(subscribersCount/getSubscribersBatchSize) + 1)`
3. `subscribers, err := c.Wallet.GetSubscribers(topic, offset, getSubscribersBatchSize, true, false)`

config.entry.json
```diff
@@ -14,5 +14,10 @@
   "ReversePrice": "0.001",
   "ReverseClaimInterval": 3600,
   "ReverseSubscriptionDuration": 180,
-  "ReverseSubscriptionFee": "0"
+  "ReverseSubscriptionFee": "0",
+  "IPFilter": {
+    "Location": {
+      "Country": "US"
```
yilunzhang, line 20: We shouldn't hard code US into the default config

tuna.go
```diff
@@ -311,6 +314,33 @@ func (c *Common) CreateServerConn(force bool) error {
...
+						_, exists := filter.WhiteList[metadata.IP]
+						if !exists {
+							continue
+						}
+					}
+					if !filter.Location.Empty() {
+						loc, err := getLocationFromIP2C(metadata.IP)
+						if err != nil {
+							log.Println(err)
+							continue RandomSubscriber
```
yilunzhang, line 334: If IP2C service is down, we should skip the filter, instead of getting stuck

tuna.go
```diff
@@ -311,6 +314,33 @@ func (c *Common) CreateServerConn(force bool) error {
...
+						}
+						if loc.Code != filter.Location.Country {
+							log.Printf("node %s from %s is non-compliant", metadata.IP, loc.Country)
```
yilunzhang, line 337: We should remove log on L337 and L340. Otherwise there will be a lot of spam logs...

util.go
```diff
@@ -1,11 +1,38 @@
...
+	Latitude  float32 `json:"Latitude"`
+	Longitude float32 `json:"Longitude"`
+}
+
+func (l *Location) Empty() bool {
+	return *l == Location{}
+}
+
+type IPFilter struct {
+	Location  *Location
```
yilunzhang, line 27: We should support:
1. a list of country code
2. country blacklist

ipFilter: {
  allow: [
    {country: "US"},
    {country: "UK"},
    {ip: "1.2.3.4"}
  ],
  disallow: [
    {country: "XX"},
    {ip: "4.3.2.1"}
  ]
}

tuna.go
```diff
@@ -126,8 +128,8 @@ func (c *Common) SetConnected(connected bool) {
 	c.connected = connected
 }
 
-func (c *Common) GetServerTCPConn(force bool) (net.Conn, error) {
-	err := c.CreateServerConn(force)
+func (c *Common) GetServerTCPConn(force bool, filter *IPFilter) (net.Conn, error) {
```
yilunzhang, line 131: No need to pass filter as a parameter as you can use `c. EntryServiceInfo.IPFilter`

tuna.go
```diff
@@ -347,11 +374,15 @@ func (c *Common) CreateServerConn(force bool) error {
 					continue RandomSubscriber
 				}
 
-				if entryToExitPrice > c.EntryToExitMaxPrice {
+				entryToExitMaxPrice, exitToEntryMaxPrice, err := ParsePrice(c.EntryServiceInfo.MaxPrice)
```
yilunzhang, line 377: It's better to parse max price in the beginning of the function so no need to parse price for many times

tuna.go
```diff
@@ -298,11 +300,18 @@ func (c *Common) CreateServerConn(force bool) error {
...
 			}
-			offset := rand.Intn(int(subscribersCount))
+			offset := rand.Intn(subscribersCount)
 			subscribers, err := c.Wallet.GetSubscribers(topic, offset, 1, true, false)
 			if err != nil {
 				return err
 			}
+			if !c.EntryServiceInfo.IPFilter.Empty() {
+				offset = rand.Intn(subscribersCount/getSubscribersBatchSize + 1)
+				subscribers, err = c.Wallet.GetSubscribers(topic, offset, offset*getSubscribersBatchSize, true, false)
```
yilunzhang, line 310: You should always do batch get subscriber even if filter is empty

geo.go
```diff
@@ -0,0 +1,133 @@
...
+
+func (l *Location) Empty() bool {
+	if l == nil {
+		return true
+	}
+	return *l == emptyLocation
+}
+
+type IPFilter struct {
+	Location *Location
```
yilunzhang, line 31: Why we still need Location here?

geo.go
```diff
@@ -0,0 +1,133 @@
...
+	Allow    map[string]interface{} `json:"Allow"`
```
yilunzhang, line 32: Shouldn't this be `[]Location`? Same for Disallow

geo.go
```diff
@@ -0,0 +1,133 @@
...
+	Disallow map[string]interface{} `json:"Disallow"`
+}
+
+func (f *IPFilter) Empty() bool {
+	if f == nil {
+		return true
+	}
+
+	return f.Location.Empty() && len(f.Allow) == 0 && len(f.Disallow) == 0
```
yilunzhang, line 41: This will never be true for the default IPFilter given in `config.entry.json`

tuna.go
```diff
@@ -311,6 +318,24 @@ func (c *Common) CreateServerConn(force bool) error {
 
 				metadata := c.GetMetadata()
 
+				if !c.EntryServiceInfo.IPFilter.Empty() {
+					if !c.EntryServiceInfo.IPFilter.IPCheck(metadata.IP) {
+						continue
+					}
+					if !c.EntryServiceInfo.IPFilter.Location.Empty() {
```
yilunzhang, line 325: These logic can be put into IPCheck() so we just need to call IPCheck() here

tuna.go
```diff
@@ -311,6 +318,24 @@ func (c *Common) CreateServerConn(force bool) error {
...
+						loc, err := getLocationFromIP2C(metadata.IP)
+						if err != nil {
+							log.Println(err)
+							c.EntryServiceInfo.IPFilter.Location = nil
```
yilunzhang, line 329: This will lead to unwanted results: if ip2c service is down for a few minutes, then recovered, it will never do ip filter again

tuna.go
```diff
@@ -311,6 +318,24 @@ func (c *Common) CreateServerConn(force bool) error {
 
 				metadata := c.GetMetadata()
 
+				if !c.EntryServiceInfo.IPFilter.Empty() {
+					if !c.EntryServiceInfo.IPFilter.IPCheck(metadata.IP) {
```
yilunzhang, line 322: We still cannot distinguish whether the ip2c service is down or returned result does not contain location (e.g. unidentified IP). One way to do this is to return (ok, err) from IPCheck. We can also put Empty() inside IPCheck, so the logic becomes:
ok, err = IPCheck()
if err == nil { if !ok then continue  }
else print error but do nothing else

tuna.go
```diff
@@ -298,8 +304,9 @@ func (c *Common) CreateServerConn(force bool) error {
 			if subscribersCount == 0 {
 				return errors.New("there is no service providers for " + c.Service.Name)
 			}
-			offset := rand.Intn(int(subscribersCount))
-			subscribers, err := c.Wallet.GetSubscribers(topic, offset, 1, true, false)
+
+			offset := rand.Intn(subscribersCount/getSubscribersBatchSize + 1)
+			subscribers, err := c.Wallet.GetSubscribers(topic, offset, offset*getSubscribersBatchSize, true, false)
```
yilunzhang, line 309: Should be `(topic, offset*getSubscribersBatchSize, getSubscribersBatchSize, true, false)`

geo.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+	}
+	for _, loc := range f.Allow {
+		if len(loc.IP) == 0 {
+			continue
+		}
+		if loc.IP == ip {
+			return true
+		}
+		return false
```
yilunzhang, line 71: Why we have `return false` here?

geo.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	return true
```
yilunzhang, line 74: If we return true at the end of the function, then f.Allow will not have any effect (only Disallow will be used).

Instead, I think we should do IPCheck and CountryCheck together for each Location.

tuna.go
```diff
@@ -74,6 +75,7 @@ type Common struct {
 	SubscriptionPrefix  string
 	Reverse             bool
 	ReverseMetadata     *Metadata
+	EntryServiceInfo    *EntryServiceInfo
```
yilunzhang, line 78: This should not be called EntryServiceInfo because it's used by both entry and exit.
Also since you use it as a replacement for `common.EntryToExitMaxPrice` and `common.ExitToEntryMaxPrice`, you should remove these two fields

tuna.go
```diff
@@ -311,6 +318,17 @@ func (c *Common) CreateServerConn(force bool) error {
...
 				metadata := c.GetMetadata()
 
+				if !c.EntryServiceInfo.IPFilter.IPCheck(metadata.IP) {
+					continue
+				}
+				loc, err := getLocationFromIP2C(metadata.IP)
+				if err != nil {
+					log.Println(err)
+				}
+				if !c.EntryServiceInfo.IPFilter.CountryCheck(loc.CountryCode) {
```
yilunzhang, line 328: This will panic if `loc == nil`

geo.go
```diff
@@ -0,0 +1,138 @@
...
+	resp, err := client.Get(queryUrl)
+	if err != nil {
+		return &Location{}, err
+	}
+
+	defer resp.Body.Close()
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
```
yilunzhang, line 116: Error here means ip2c service is down or network problem. It does not mean the input is invalid

geo.go
```diff
@@ -0,0 +1,130 @@
...
+	loc, err := getLocationFromIP2C(ip)
+	if err != nil {
+		return false, err
+	}
+
+	valid := ValidCheck(f, loc)
+	return valid, nil
+}
+
+func ValidCheck(f *IPFilter, loc *Location) bool {
```
yilunzhang, line 115: This cannot pass check if
`f.Allow == [emptyLocation]`
and
`f.Disallow == [emptyLocation]`

tests/geo_test.go
```diff
@@ -0,0 +1,44 @@
...
+	{
+		f: tuna.IPFilter{
+			Disallow: []tuna.Location{{IP: "1.2.3.4"}},
+		},
+		location: tuna.Location{IP: "1.2.3.4"},
+		result:   false,
+	},
+}
+
+func TestAllowLocation(t *testing.T) {
```
yilunzhang, line 32: `TestAllowLocation` and `TestDisAllowLocation` should be combined into one function, which loop over all testData and compare results

tuna.go
```diff
@@ -65,15 +72,14 @@ type Metadata struct {
...
-	ListenIP            net.IP
-	EntryToExitMaxPrice common.Fixed64
-	ExitToEntryMaxPrice common.Fixed64
-	Wallet              *nkn.Wallet
-	DialTimeout         uint16
-	SubscriptionPrefix  string
-	Reverse             bool
-	ReverseMetadata     *Metadata
+	Service            *Service
+	ListenIP           net.IP
```
yilunzhang, line 76: ListenIP should be removed too since it's contained in ServiceInfo

geo.go
```diff
@@ -0,0 +1,130 @@
...
+
+func (l *Location) Empty() bool {
+	if l == nil {
+		return true
+	}
+	return *l == emptyLocation
+}
+
+func (l *Location) Match(location *Location) bool {
+	if len(location.IP) > 0 && location.IP == l.IP {
```
yilunzhang, line 30: Should this be `len(location.IP) > 0` or `len(l.IP) > 0`?

exit.go
```diff
@@ -376,19 +375,13 @@ func (te *TunaExit) StartReverse(serviceName string) error {
...
-		Service:             &Service{Name: DefaultReverseServiceName},
-		EntryToExitMaxPrice: maxPrice,
-		ExitToEntryMaxPrice: maxPrice,
-		Wallet:              te.Wallet,
-		DialTimeout:         te.config.DialTimeout,
-		ReverseMetadata:     reverseMetadata,
-		SubscriptionPrefix:  te.config.SubscriptionPrefix,
+		Service:            &Service{Name: DefaultReverseServiceName},
+		Wallet:             te.Wallet,
+		ServiceInfo:        &ServiceInfo{MaxPrice: te.config.ReverseMaxPrice},
```
yilunzhang, line 381: Did you miss to pass reverse ip filter?

geo.go
```diff
@@ -0,0 +1,130 @@
...
+	loc, err := parseIP2C(string(body))
+	if err != nil {
+		return unknown, err
+	}
+	loc.IP = ip
+	return loc, nil
+}
+
+func parseIP2C(body string) (*Location, error) {
+	if body[0:1] != "1" {
```
yilunzhang, line 88: 1. You need to handle `len(body) == 0` to avoid panic
2. `body[0]` is enough instead of `body[0:1]`

geo.go
```diff
@@ -0,0 +1,130 @@
...
+	l.CountryCode = res[1]
+	l.Country = res[3]
+	return l, nil
+}
+
+func geoCheck(f *IPFilter, ip string) (bool, error) {
+	if f.Empty() {
+		return true, nil
+	}
+	loc, err := getLocationFromIP2C(ip)
```
yilunzhang, line 106: If ip2c service is down, tuna will not be available using the current logic. Please check VERY CAREFULLY and add test case for it as well.

geo.go
```diff
@@ -0,0 +1,130 @@
...
+
+func getLocationFromIP2C(ip string) (*Location, error) {
+	queryUrl := IP2CUrl + ip
+	client := http.Client{
+		Timeout: 10 * time.Second,
+	}
+	unknown := &Location{CountryCode: "UNKNOWN"}
+	resp, err := client.Get(queryUrl)
+	if err != nil {
+		return unknown, err
```
yilunzhang, line 70: Should pass check

geo.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	defer resp.Body.Close()
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return unknown, err
```
yilunzhang, line 77: Should pass check

geo.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	loc, err := parseIP2C(string(body))
+	if err != nil {
+		return unknown, err
```
yilunzhang, line 81: Should NOT pass check

geo.go
bufrr: body[0] will return uint8 rather than string, got this hint from IDE

yilunzhang: OK

config.entry.json
```diff
@@ -1,7 +1,13 @@
 {
   "Services": {
     "httpproxy": {
-      "maxPrice": "0.001"
+      "maxPrice": "0.001",
+      "IPFilter": {
+        "Allow": [
```
yilunzhang, line 6: As we discussed, we need to add a Location with empty country in default config so user will know what to add

geo.go
yilunzhang: The current code still cannot pass the above example

tests/geo_test.go
```diff
@@ -0,0 +1,105 @@
...
+		location: tuna.Location{IP: IP1},
+		result:   true,
+	},
+	{
+		f: tuna.IPFilter{
+			Allow:    []tuna.Location{{IP: IP1}},
+			Disallow: []tuna.Location{{IP: IP1}},
+		},
+		location: tuna.Location{IP: IP1},
+		result:   true,
```
yilunzhang, line 49: Disallow should have priority over Allow and the expected result should be false

tests/geo_test.go
```diff
@@ -0,0 +1,105 @@
...
+		result:   false,
+	},
+	{
+		f: tuna.IPFilter{
+			Allow: []tuna.Location{{IP: IP1}, {IP: IP2}},
+		},
+		location: tuna.Location{IP: IP3},
+		result:   false,
+	},
+}
```
yilunzhang, line 96: You also need to test CountryCode

tests/geo_test.go
```diff
@@ -0,0 +1,229 @@
...
+		result:   true,
+	},
+	{
+		f: tuna.IPFilter{
+			Allow: []tuna.Location{{CountryCode: "US"}},
+		},
+		location: tuna.Location{CountryCode: "CA"},
+		result:   false,
+	},
+	{
```
yilunzhang, line 176: duplicate

tests/geo_test.go
```diff
@@ -0,0 +1,229 @@
...
+		result:   true,
+	},
+	{
+		f: tuna.IPFilter{
+			Disallow: []tuna.Location{{IP: IP3}},
+		},
+		location: tuna.Location{IP: IP4},
+		result:   true,
+	},
+	{
```
yilunzhang, line 90: Duplicate (L45-51)

PR #8, opened 2020-04-22
Remove unexpected entry close call

bufrr:


config.exit.json
```diff
@@ -4,7 +4,7 @@
   "ListenUDP": 30011,
   "Reverse": false,
   "ReverseRandomPorts": true,
-  "ReverseMaxPrice": "0",
+  "ReverseMaxPrice": "0.001",
```
yilunzhang, line 7: It's better to use 0 as default price so people who just generate wallet can try with the free nodes we provide

entry.go
```diff
@@ -219,7 +219,6 @@ func (te *TunaEntry) StartReverse(stream *smux.Stream) error {
 			stream, err = session.AcceptStream()
 			if err != nil {
 				log.Println("Close connection:", err)
-				te.close()
```
yilunzhang, line 222: Why not close te here? When accept returns error, it means the connection is already broken and cannot be used anymore.

bufrr: Entry will closed at L308, this line will cause panic if exit close.The code in master branch doesn't call `te.close` either.

yilunzhang: A more robust solution is to allow close() to be called multiple times. Example: https://github.com/nknorg/nkn-sdk-go/blob/master/client.go#L168

config.exit.json
bufrr: The default price in entry config is 0.001, shoule we change the price to 0 together?

yilunzhang: We don't need to change provider side because provider wants to earn money (default price > 0), while user wants to use (default price = 0)

yilunzhang: Good point, it's better to keep them the same

yilunzhang: Let's keep 0.001 for both of them then

entry.go
yilunzhang: We need to add back te.close() here

PR #16, opened 2020-05-12
retry when get err from ip2c api

bufrr:


geo.go
```diff
@@ -108,7 +118,7 @@ func (f *IPFilter) GeoCheck(ip string) (bool, error) {
 	if f.Empty() {
 		return true, nil
 	}
-	loc, err := getLocationFromIP2C(ip)
+	loc, err := getLocationFromIP2C(ip, 3)
```
yilunzhang, line 121: It's better to make 3 a constant

geo.go
```diff
@@ -60,18 +60,28 @@ func (f *IPFilter) Empty() bool {
...
+	i := 0
+	var resp *http.Response
+	var err error
+	for i < retry {
+		resp, err = client.Get(queryUrl)
+		if err != nil {
+			i++
+			log.Println(err)
+			continue
+		} else if i == retry-1 {
```
yilunzhang, line 78: The `for` loop will never end if `err != nil` is always true

geo.go
```diff
@@ -60,18 +60,28 @@ func (f *IPFilter) Empty() bool {
...
+			return &emptyLocation, err
+		} else {
+			defer resp.Body.Close()
+			break
+		}
+	}
```
yilunzhang, line 84: A generic way to do retry:
```
i := 0
for ; i < retry; i++ {
    err = xxx
    if err != nil {
        // do something on error like log
        continue
    }
    break
}
if i == retry {
    // on failure
    return
}

// on success
```

entry.go
```diff
@@ -194,6 +194,7 @@ func (te *TunaEntry) StartReverse(stream *smux.Stream) error {
 		claimInterval := time.Duration(te.config.ReverseClaimInterval) * time.Second
 		onErr := nkn.NewOnError(1, nil)
 		isClosed := false
+		reversePrice, _ := common.StringToFixed64(te.config.ReversePrice)
```
yilunzhang, line 197: 1. should use `ParsePrice` instead of `common.StringToFixed64`
2. never ignore error

PR #20, opened 2020-05-28
add traffic charging calcalation and remove traffic coverage verify

bufrr:


config.entry.json
```diff
@@ -22,5 +22,6 @@
   "reversePrice": "0.001",
   "reverseClaimInterval": 3600,
   "reverseSubscriptionDuration": 180,
-  "reverseSubscriptionFee": "0"
+  "reverseSubscriptionFee": "0",
+  "minChargingUnit": 100
```
yilunzhang, line 26: This shouldn't be configurable, because mismatch between entry and exit will cause service provider to close connection

entry.go
```diff
@@ -56,13 +57,13 @@ type TunaEntry struct {
 }
 
 func NewTunaEntry(service *Service, serviceInfo *ServiceInfo, config *EntryConfiguration, wallet *nkn.Wallet) (*TunaEntry, error) {
-	common, err := NewCommon(service, serviceInfo, wallet, config.DialTimeout, config.SubscriptionPrefix, config.Reverse, nil)
+	newCommon, err := NewCommon(service, serviceInfo, wallet, config.DialTimeout, config.SubscriptionPrefix, config.Reverse, nil)
```
yilunzhang, line 60: Why rename?

exit.go
```diff
@@ -63,13 +64,13 @@ type TunaExit struct {
 }
 
 func NewTunaExit(config *ExitConfiguration, services []Service, wallet *nkn.Wallet) (*TunaExit, error) {
-	common, err := NewCommon(nil, nil, wallet, config.DialTimeout, config.SubscriptionPrefix, config.Reverse, nil)
+	newCommon, err := NewCommon(nil, nil, wallet, config.DialTimeout, config.SubscriptionPrefix, config.Reverse, nil)
```
yilunzhang, line 67: Why rename?

exit.go
```diff
@@ -128,10 +129,22 @@ func (te *TunaExit) handleSession(session *smux.Session) {
 		if err != nil {
 			log.Fatalln(err)
 		}
+		getTotalTraffic := func() common.Fixed64 {
+			totalTraffic := common.Fixed64(0)
+			for i := range bytesIn {
+				in := atomic.LoadUint64(&bytesIn[i])
+				out := atomic.LoadUint64(&bytesOut[i])
+				if in == 0 && out == 0 {
```
yilunzhang, line 137: This if condition is not needed here

tuna.go
```diff
@@ -494,59 +494,83 @@ func (c *Common) CreateServerConn(force bool) error {
...
+			}
+		}
+	}()
+
+	go func() {
+	loop:
+		for {
+			select {
+			case <-resetChan:
+				goto loop
```
yilunzhang, line 527: This `goto` is not doing anything

tuna.go
```diff
@@ -494,59 +494,83 @@ func (c *Common) CreateServerConn(force bool) error {
...
-
-		bytesIn := atomic.LoadUint64(bytesInUsed)
-		bytesOut := atomic.LoadUint64(bytesOutUsed)
-		cost := exitToEntryPrice*common.Fixed64(bytesIn-*bytesInPaid)/TrafficUnit + entryToExitPrice*common.Fixed64(bytesOut-*bytesOutPaid)/TrafficUnit
-		if cost == lastCost {
-			time.Sleep(DefaultNanoPayUpdateInterval / 10)
-			continue
-		}
-		lastCost = cost
+		case cost := <-costChan:
```
yilunzhang, line 548: This loop can be combined with two goroutines before:
```
paymentTimer = new timer (DefaultNanoPayUpdateInterval)
for {
    select {
    case <- time.After(100ms):
        continue if delta_bytes < threshold
    case <- paymentTimer:
    }
    // pay
}
```

tuna.go
```diff
@@ -494,59 +494,83 @@ func (c *Common) CreateServerConn(force bool) error {
...
 		}
-
-		err = sendNanoPay(np, paymentStream, cost)
-		if err != nil {
-			log.Printf("send nano payment err: %v", err)
-			time.Sleep(time.Second)
-			continue
-		}
-
-		*bytesInPaid = bytesIn
```
yilunzhang, line 546: You shouldn't remove this. `bytesInPaid` and `bytesOutPaid` should ONLY be updated if send nanopay success

tuna.go
```diff
@@ -832,6 +858,16 @@ func checkPaymentTimeout(session *smux.Session, updateTimeout time.Duration, las
 		}
 		lastCost = totalCost
 
+		totalTraffic = getTotalTraffic()
```
yilunzhang, line 861: The logic is not right here. If there are some small traffic after tuna conn is established, the service provider side will disconnect after 2s because user side has not pay anything yet. Have you really tested it yet?

config.entry.json
yilunzhang: Also the meaning of `minChargingUnit` is not clear. Suggest to change to `DefaultNanoPayUpdateThreshold`

tuna.go
```diff
@@ -871,13 +907,10 @@ func writeStreamMetadata(stream *smux.Stream, streamMetadata *pb.StreamMetadata)
 }
 
 func handlePaymentStream(
```
yilunzhang, line 909: You shouldn't change this function

tuna.go
yilunzhang: It's better not to change `checkPaymentTimeout`, but create another function:
```
for every second {
    if totalTraffic - lastPaidTraffic > threshold / MinTrafficCoverage {
        // fail to pay logic
    }
}
```

entry.go
bufrr: same name with import package, got this hint from IDE

tuna.go
```diff
@@ -503,20 +504,27 @@ func (c *Common) startPayment(
...
+	paymentTimer := time.After(DefaultNanoPayUpdateInterval)
 	for {
 		select {
 		case <-closeChan:
 			return
-		default:
+		case <-time.After(100 * time.Millisecond):
+			bytesIn = atomic.LoadUint64(bytesInUsed)
+			bytesOut = atomic.LoadUint64(bytesOutUsed)
+			if (bytesIn+bytesOut)-(*bytesInPaid+*bytesOutPaid) <= uint64(20*TrafficUnit) {
```
yilunzhang, line 519: 20 should be a constant

tuna.go
```diff
@@ -503,20 +504,27 @@ func (c *Common) startPayment(
...
+				continue
+			}
+			cost = exitToEntryPrice*common.Fixed64(bytesIn-*bytesInPaid)/TrafficUnit + entryToExitPrice*common.Fixed64(bytesOut-*bytesOutPaid)/TrafficUnit
```
yilunzhang, line 522: This line is duplicate of L527 and should be removed

tuna.go
```diff
@@ -503,20 +504,27 @@ func (c *Common) startPayment(
...
+		case <-paymentTimer:
+			paymentTimer = time.After(2 * time.Second)
 		}
 
-		bytesIn := atomic.LoadUint64(bytesInUsed)
-		bytesOut := atomic.LoadUint64(bytesOutUsed)
-		cost := exitToEntryPrice*common.Fixed64(bytesIn-*bytesInPaid)/TrafficUnit + entryToExitPrice*common.Fixed64(bytesOut-*bytesOutPaid)/TrafficUnit
+		cost = exitToEntryPrice*common.Fixed64(bytesIn-*bytesInPaid)/TrafficUnit + entryToExitPrice*common.Fixed64(bytesOut-*bytesOutPaid)/TrafficUnit
```
yilunzhang, line 527: L517 and L518 needs to be copied to here also

tuna.go
```diff
@@ -845,6 +851,25 @@ func checkPaymentTimeout(session *smux.Session, updateTimeout time.Duration, las
 	}
 }
 
+func checkBytesOwned(session *smux.Session, isClosed *bool, bytesPaid *common.Fixed64, getTotalTraffic func() common.Fixed64) {
```
yilunzhang, line 854: Suggest rename to `checkTrafficPaid`

exit.go
```diff
@@ -151,19 +153,31 @@ func (te *TunaExit) handleSession(session *smux.Session) {
...
-		return cost
+		return cost, totalBytes
 	}
 
 	if !te.config.Reverse {
 		npc, err = te.Wallet.NewNanoPayClaimer(te.config.BeneficiaryAddr, int32(claimInterval/time.Millisecond), onErr)
 		if err != nil {
 			log.Fatalln(err)
 		}
+		getTotalTraffic := func() common.Fixed64 {
```
yilunzhang, line 166: Why not use `getTotalCost` as `getTotalTraffic` since it also returns total traffic?

entry.go
```diff
@@ -179,16 +180,25 @@ func (te *TunaEntry) StartReverse(stream *smux.Stream) error {
 		return err
 	}
 
-	getTotalCost := func() common.Fixed64 {
+	getTotalCost := func() (common.Fixed64, common.Fixed64) {
+		in := common.Fixed64(atomic.LoadUint64(&te.reverseBytesIn))
+		out := common.Fixed64(atomic.LoadUint64(&te.reverseBytesOut))
+		time.Sleep(TrafficCalDelay * time.Second)
```
yilunzhang, line 186: We shouldn't add delay here. Instead, it should be added to caller of `getTotalCost`

tuna.go
```diff
@@ -845,6 +851,25 @@ func checkPaymentTimeout(session *smux.Session, updateTimeout time.Duration, las
...
+	for {
+		before := getTotalTraffic()
+		time.Sleep(2 * time.Second)
```
yilunzhang, line 857: Should we use `TrafficCalDelay` constant?

tuna.go
```diff
@@ -45,6 +45,7 @@ const (
 	DefaultServiceListenIP                 = "127.0.0.1"
 	DefaultReverseServiceListenIP          = "0.0.0.0"
 	TrafficUnit                            = 1024 * 1024
+	TrafficCalDelay                        = 5
```
yilunzhang, line 48: Suggest to increase it to `10 * time.Second`

tuna.go
```diff
@@ -845,6 +851,25 @@ func checkPaymentTimeout(session *smux.Session, updateTimeout time.Duration, las
...
+	for {
+		before := getTotalTraffic()
+		time.Sleep(2 * time.Second)
+		after := getTotalTraffic()
+		if *isClosed {
+			break
+		}
+
+		unpaidTraffic := before - *bytesPaid
+		if unpaidTraffic > 30*(after-before) && unpaidTraffic > common.Fixed64(60*TrafficUnit) {
```
yilunzhang, line 864: Why we need `unpaidTraffic > 30*(after-before)`?

tuna.go
```diff
@@ -879,29 +904,33 @@ func handlePaymentStream(
...
+		*bytesPaid = totalBytes
+		now := time.Now()
+		diff := now.Sub(*lastUpdate)
+		*lastUpdate = now
 
-		totalCost := getTotalCost()
-		*lastUpdate = time.Now()
-
-		if float64(*lastClaimed) < float64(totalCost)*MinTrafficCoverage {
+		if float64(*lastClaimed) < float64(totalCost)*MinTrafficCoverage && diff.Seconds() > 30 {
```
yilunzhang, line 930: This check can be bypassed as longer as user pay a tiny amount (e.g 1% of cost) every 30s

tuna.go
```diff
@@ -40,12 +40,14 @@ const (
 	UDP                           Protocol = "udp"
 	DefaultNanoPayDuration                 = 4320 * 30
 	DefaultNanoPayUpdateInterval           = time.Minute
-	DefaultSubscriptionPrefix              = "tuna_v1."
+	DefaultSubscriptionPrefix              = "tuna_v12345."
```
yilunzhang, line 43: You forget to change this

PR #43, opened 2020-09-22
Add geo support for AWS & GCP when using IPFilter

bufrr:


geo/aws.go
```diff
@@ -0,0 +1,121 @@
...
+	AWSGeoUrl  = "https://ip-ranges.amazonaws.com/ip-ranges.json"
+	AWSGeoFile = "ip-ranges.json"
+)
+
+type AWSProvider struct {
+}
+
+type AWSGeoInfo struct {
+	SyncToken  string      `json:"syncToken"`
+	CreateData string      `json:"createDate"`
```
yilunzhang, line 22: This should be `CreateDate` instead of `CreateData`

geo/aws.go
```diff
@@ -0,0 +1,121 @@
...
+var AWSRegionMapping = map[string]string{
+	"us-east-1":      "US",
+	"us-east-2":      "US",
+	"us-west-1":      "US",
+	"us-west-2":      "US",
+	"af-south-1":     "ZA",
+	"ap-east-1":      "HK",
+	"ap-south-1":     "IN",
+	"ap-northeast-1": "JP",
+	"ap-northeast-2": "KR",
```
yilunzhang, line 42: I think you are missing ap-northeast-3: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html

geo/aws.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+		defer resp.Body.Close()
+
+		_, err = io.Copy(AWSFile, resp.Body)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	AWSFile, err := os.Open(AWSGeoFile)
```
yilunzhang, line 84: You are reading the whole data from disk for every request

geo/aws.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+		defer AWSFile.Close()
+
+		resp, err := http.Get(AWSGeoUrl)
+		if err != nil {
+			return nil, err
+		}
+		defer resp.Body.Close()
+
+		_, err = io.Copy(AWSFile, resp.Body)
```
yilunzhang, line 78: It's better to fetch it to memory, make sure data is valid, and then write to disk. Otherwise you might write partial or invalid data to disk

geo/aws.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func parseAWS(ip string, info *AWSGeoInfo) *Location {
+	loc := Location{}
+	parsed := net.ParseIP(ip)
+	for _, p := range info.Prefixes {
+		if len(p.IPPrefix) == 0 {
+			continue
+		}
+		_, subnet, _ := net.ParseCIDR(p.IPPrefix)
```
yilunzhang, line 111: This should only be done once, not every time there is a request

geo/aws.go
```diff
@@ -0,0 +1,121 @@
...
+	"ap-east-1":      "HK",
+	"ap-south-1":     "IN",
+	"ap-northeast-1": "JP",
+	"ap-northeast-2": "KR",
+	"ap-southeast-1": "SG",
+	"ap-southeast-2": "AU",
+	"ca-central-1":   "CA",
+	"eu-central-1":   "DE",
+	"eu-west-1":      "IE",
+	"eu-west-2":      "EN",
```
yilunzhang, line 48: Where did you get `EN`? It should be `GB`: https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes#UNI2

yilunzhang: Also please double check other codes to make sure there are no mistakes

geo/aws.go
```diff
@@ -0,0 +1,121 @@
...
+func (p AWSProvider) GetLocation(ip string) (*Location, error) {
+	loc, err := getLocationFromAWS(ip)
+	if err != nil {
+		return &emptyLocation, err
+	}
+	return loc, nil
+}
+
+func getLocationFromAWS(ip string) (*Location, error) {
+	if _, err := os.Stat(AWSGeoFile); os.IsNotExist(err) {
```
yilunzhang, line 65: File content should be updated regularly

geo/geo.go
```diff
@@ -121,62 +113,26 @@ func (f *IPFilter) AllowLocation(loc *Location) bool {
...
-	if body[0] != byte('1') {
-		return nil, errors.New("get ip2c result err")
-	}
-
-	res := strings.Split(body, ";")
-	if len(res) != 4 {
-		return nil, errors.New("invalid response from ip2c service")
+func getLocation(ip string) *Location {
+	providers := []interface{}{AWSProvider{}, GCPProvider{}, MaxMindProvider{}, IP2CProvider{}}
+	//providers := []interface{}{MaxMindProvider{}}
```
yilunzhang, line 126: Remove test code

geo/ip2c.go
```diff
@@ -0,0 +1,82 @@
...
+	"io/ioutil"
+	"log"
+	"net/http"
+	"strings"
+	"time"
+)
+
+const (
+	GeoIPRetry = 3
+	IP2CUrl    = "https://ip2c.nkn.org/"
```
yilunzhang, line 14: Let's use `http://ip2c.org` as default

tests/geo_test.go
```diff
@@ -3,208 +3,248 @@ package tests
...
+		IP:      "52.194.247.186", // AWS
+		Country: "JP",
+	},
+	{
+		IP:      "144.91.100.54", // Contabo
+		Country: "DE",
+	},
+	{
+		IP:      "3.9.118.165", // AWS
+		Country: "EN",
```
yilunzhang, line 244: Also this one

geo/gcp.go
```diff
@@ -0,0 +1,137 @@
...
+	"asia-northeast1":         "JP",
+	"asia-northeast2":         "JP",
+	"asia-northeast3":         "KR",
+	"asia-south1":             "IN",
+	"asia-southeast1":         "SG",
+	"asia-southeast2":         "ID",
+	"australia-southeast1":    "AU",
+	"europe-north1":           "FI",
+	"europe-west1":            "BE",
+	"europe-west2":            "EN",
```
yilunzhang, line 48: This is still wrong

geo/geo.go
```diff
@@ -121,62 +155,44 @@ func (f *IPFilter) AllowLocation(loc *Location) bool {
...
+}
 
-	i := 0
-	var resp *http.Response
-	var err error
-	for ; i < retry; i++ {
-		resp, err = client.Get(queryURL)
+func GetJsonFile(url, filename string) ([]byte, error) {
+	if _, err := os.Stat(filename); os.IsNotExist(err) {
+		AWSFile, err := os.Create(filename)
```
yilunzhang, line 168: Name should not be AWSFile

geo/geo.go
```diff
@@ -121,62 +155,44 @@ func (f *IPFilter) AllowLocation(loc *Location) bool {
...
-		return &emptyLocation, err
-	}
+		defer AWSFile.Close()
 
-	defer resp.Body.Close()
-
-	body, err := ioutil.ReadAll(resp.Body)
-	if err != nil {
-		return &emptyLocation, err
+		resp, err := http.Get(url)
```
yilunzhang, line 174: Never use http.get, always creates a `http.Client` and set a timeout, otherwise this function may stuck forever. You can use a long timeout like 5 or 10 minutes.

geo/maxmind.go
```diff
@@ -0,0 +1,87 @@
...
+		return nil, errors.New("invalid DBFile path")
+	}
+	if _, err := os.Stat(DBFile); os.IsNotExist(err) {
+		MMFile, err := os.Create(DBFile)
+		if err != nil {
+			return nil, err
+		}
+		defer MMFile.Close()
+
+		resp, err := http.Get(MaxMindUrl)
```
yilunzhang, line 56: Never use http.get, always creates a `http.Client` and set a timeout, otherwise this function may stuck forever. You can use a very long timeout like 30 or 60 minutes.

geo/aws.go
```diff
@@ -0,0 +1,135 @@
...
+	aws.Info = info
+	aws.InfoFile = AWSFile
+	aws.LastUpdate = time.Now()
+
+	return aws, nil
+}
+
+func (p *AWSProvider) GetLocation(ip string) (*Location, error) {
+	if time.Since(p.LastUpdate) > AWSExpired {
+		err := os.Remove(p.InfoFile)
```
yilunzhang, line 96: Remove the old file before upgrade success is not a good way. We need to make sure if upgrade fails, old file can still be used.

geo/geo.go
```diff
@@ -40,8 +41,12 @@ func (l *Location) Match(location *Location) bool {
 }
 
 type IPFilter struct {
-	Allow    []Location `json:"allow"`
-	Disallow []Location `json:"disallow"`
+	Allow       []Location    `json:"allow"`
+	Disallow    []Location    `json:"disallow"`
+	AWSFile     string        `json:"awsfile"`
```
yilunzhang, line 46: It's better to let user configure a directory and put all geo files there

geo/ip2c.go
```diff
@@ -0,0 +1,86 @@
...
+	"io/ioutil"
+	"log"
+	"net/http"
+	"strings"
+	"time"
+)
+
+const (
+	GeoIPRetry = 3
+	IP2CUrl    = "http://ip2c.nkn.org/"
```
yilunzhang, line 14: Should be `ip2c.org`, not `ip2c.nkn.org` (which does not support http)

geo/geo.go
```diff
@@ -84,19 +89,48 @@ func (f *IPFilter) AllowIP(ip string) (bool, error) {
...
+		loc = f.GetLocation(ip)
 	} else {
 		loc = &Location{IP: ip}
 	}
 
 	return f.AllowLocation(loc), nil
 }
 
+func (f *IPFilter) GetLocation(ip string) *Location {
+	if len(f.Providers) == 0 {
```
yilunzhang, line 102: You need to consider the case that GetLocation is called concurrently.

geo/geo.go
bufrr: User can config geo files in config files in `IPFilter` scope

geo/geo.go
```diff
@@ -0,0 +1,231 @@
...
+			Timeout: 5 * time.Second,
+		}
+		resp, err := client.Get(url)
+		if err != nil {
+			return nil, err
+		}
+		defer resp.Body.Close()
+		b, err := ioutil.ReadAll(resp.Body)
+		if !json.Valid(b) {
+			os.Remove(filename)
```
yilunzhang, line 171: You can create file after json.Valid() so you don't have to remove file here

geo/geo.go
```diff
@@ -0,0 +1,231 @@
...
+			return nil, errors.New("invalid json")
+		}
+		_, err = f.Write(b)
+		if err != nil {
+			return nil, err
```
yilunzhang, line 176: Should remove file here

geo/aws.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+func (p *AWSProvider) Init() {
+	b, err := GetJsonFile(p.url, p.fileName)
+	if err != nil {
+		log.Print(err)
+	}
+	err = json.Unmarshal(b, &p.Info)
+	if err != nil {
+		log.Print(err)
```
yilunzhang, line 77: Should return error

geo/aws.go
```diff
@@ -0,0 +1,160 @@
...
+	"net"
+	"os"
+	"path/filepath"
+	"time"
+)
+
+const (
+	AWSGeoUrl  = "https://ip-ranges.amazonaws.com/ip-ranges.json"
+	AWSExpired = 7 * 24 * time.Hour
+	AWSFile    = "ip-ranges.json"
```
yilunzhang, line 15: sugget name: `aws-ip.json`

geo/gcp.go
```diff
@@ -0,0 +1,162 @@
...
+	"net"
+	"os"
+	"path/filepath"
+	"time"
+)
+
+const (
+	GCPGeoUrl  = "https://www.gstatic.com/ipranges/cloud.json"
+	GCPExpired = 7 * 24 * time.Hour
+	GCPFile    = "cloud.json"
```
yilunzhang, line 15: sugget name: `gcp-ip.json`

geo/maxmind.go
```diff
@@ -0,0 +1,122 @@
...
+	"path/filepath"
+	"time"
+
+	"github.com/oschwald/geoip2-golang"
+)
+
+const (
+	MaxMindUrl     = "https://raw.githubusercontent.com/leo108/geolite2-db/master/Country.mmdb"
+	MaxMindExpired = 30 * 24 * time.Hour
+	MaxMindFile    = "Country.mmdb"
```
yilunzhang, line 18: sugget name: `geolite2-country.mmdb`

geo/maxmind.go
```diff
@@ -0,0 +1,122 @@
...
+	"net/http"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/oschwald/geoip2-golang"
+)
+
+const (
+	MaxMindUrl     = "https://raw.githubusercontent.com/leo108/geolite2-db/master/Country.mmdb"
```
yilunzhang, line 16: The name of this db is geolite2, not maxmind (maxmind has other geoip db like geoip2)

geo/geo.go
```diff
@@ -0,0 +1,231 @@
...
+		ip2c := NewIP2CProvider()
+		ip2c.Init()
+		f.Providers = []GeoProvider{aws, gcp, mm, ip2c}
+	}
+}
+
+func (f *IPFilter) UpdateJsonFile() {
+	for {
+		time.Sleep(1 * time.Hour)
+		for idx, p := range f.Providers[:3] {
```
yilunzhang, line 208: Do not hard code 3 here. You can use `len(p.Filename()) > 0` or `p.Expire() > 0` to determine whether a provider needs update

tuna.go
```diff
@@ -465,6 +467,10 @@ func (c *Common) CreateServerConn(force bool) error {
 				allSubscribers[i], allSubscribers[j] = allSubscribers[j], allSubscribers[i]
 			})
 
+			if c.ServiceInfo.IPFilter.NeedGeoInfo() {
```
yilunzhang, line 470: This will be called for EVERY new connection

tuna.go
```diff
@@ -465,6 +467,10 @@ func (c *Common) CreateServerConn(force bool) error {
...
+				c.ServiceInfo.IPFilter.AddProvider()
+				go c.ServiceInfo.IPFilter.UpdateJsonFile()
```
yilunzhang, line 472: You need to make sure goroutine exits when tuna instance stops

geo/geo.go
```diff
@@ -0,0 +1,231 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return b, nil
+}
+
+func (f *IPFilter) AddProvider() {
+	if len(f.Providers) == 0 {
+		aws := NewAWSProvider(f.DBPath)
+		aws.Init()
```
yilunzhang, line 194: You are forcing tuna to download db on every start. It's better to change Init() to MaybeUpdate() and check expiration in the function.

geo/geo.go
```diff
@@ -0,0 +1,231 @@
...
+			p.SetFileName(tmpFile.Name())
+			p.Init()
+
+			err = os.Rename(tmpFile.Name(), filename)
+			if err != nil {
+				log.Print(err)
+				os.Remove(tmpFile.Name())
+				continue
+			}
+			f.Providers[idx] = p
```
yilunzhang, line 228: You should call rename again after file move

geo/geo.go
```diff
@@ -0,0 +1,231 @@
...
+	GetLocation(ip string) (*Location, error)
+	FileName() string
+	DownloadUrl() string
+	LastUpdate() time.Time
+	Expire() time.Duration
+	Init()
+	Ready() bool
+	SetReady(bool)
+	SetFileName(string)
+	Update()
```
yilunzhang, line 24: This function is no longer needed

geo/geo.go
```diff
@@ -0,0 +1,231 @@
...
+		gcp.Init()
+		mm := NewMaxMindProvider(f.DBPath)
+		mm.Init()
+		ip2c := NewIP2CProvider()
+		ip2c.Init()
+		f.Providers = []GeoProvider{aws, gcp, mm, ip2c}
+	}
+}
+
+func (f *IPFilter) UpdateJsonFile() {
```
yilunzhang, line 205: Each service will create and start a new tuna, you need to make sure if multiple tuna instances are using the same file, they don't update it multiple times

geo/geo.go
```diff
@@ -0,0 +1,243 @@
...
+	loc, err := p.GetLocation(ip)
+	if err != nil {
+		log.Println(err)
+	}
+	return *loc
+}
+
+func GetJsonFile(url, filename string) ([]byte, error) {
+	if s, err := os.Stat(filename); os.IsNotExist(err) || s.Size() == 0 {
+		f, err := os.Create(filename)
```
yilunzhang, line 155: You need to make sure the file is removed if ANY error occurs before L177, or create file just before write.

geo/geo.go
```diff
@@ -0,0 +1,243 @@
...
+				continue
+			}
+			err = os.Rename(tmpFile.Name(), filename)
+			if err != nil {
+				log.Print(err)
+				os.Remove(tmpFile.Name())
+				continue
+			}
+			p.SetFileName(filename)
+			f.Providers[idx] = p
```
yilunzhang, line 240: It's better to use `Providers []*GeoProvider` instead of `Providers []GeoProvider` to avoid this

geo/geo.go
yilunzhang, line 155: You need to remove the file for ANY error that occurs before L177, or you should create the file just before writing to file

geo/geo.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+	if len(l.CountryCode) > 0 && strings.ToLower(location.CountryCode) == strings.ToLower(l.CountryCode) {
+		return true
+	}
+	return false
+}
+
+type IPFilter struct {
+	DBPath    string        `json:"dbpath"`
+	Download  bool          `json:"download"`
```
yilunzhang, line 54: 1. Suggest rename to 'DownloadDB'
2. `DBPath` and `Download` should be configured per tuna, not per service (per IPFilter), same for `Providers`.

geo/geo.go
```diff
@@ -0,0 +1,243 @@
...
+		return nil, err
+	}
+	return b, nil
+}
+
+func (f *IPFilter) AddProvider() {
+	if len(f.Providers) == 0 {
+		if f.Download {
+			aws := NewAWSProvider(f.DBPath)
+			aws.MaybeUpdate()
```
yilunzhang, line 195: Should print or handle error, same for other `MaybeUpdate`

geo/geo.go
```diff
@@ -0,0 +1,243 @@
...
+			gcp := NewGCPProvider(f.DBPath)
+			gcp.MaybeUpdate()
+			mm := NewMaxMindProvider(f.DBPath)
+			mm.MaybeUpdate()
+			f.Providers = append(f.Providers, aws, gcp, mm)
```
yilunzhang, line 200: Should only append a provider if maybeupdate does not return error

geo/geo.go
```diff
@@ -0,0 +1,243 @@
...
+		}
+
+		ip2c := NewIP2CProvider()
+		ip2c.MaybeUpdate()
```
yilunzhang, line 204: Same here

geo/geo.go
```diff
@@ -0,0 +1,243 @@
...
+			tmpFile, err := ioutil.TempFile("", filename+"-*")
+			if err != nil {
+				log.Print(err)
+			}
+
+			p.SetFileName(tmpFile.Name())
+			err = p.MaybeUpdate()
+			if err != nil {
+				log.Print(err)
+				continue
```
yilunzhang, line 231: Should set back to original file if error

geo/geo.go
```diff
@@ -0,0 +1,243 @@
...
+			}
+			err = os.Rename(tmpFile.Name(), filename)
+			if err != nil {
+				log.Print(err)
+				os.Remove(tmpFile.Name())
+				continue
```
yilunzhang, line 237: Should set back to original file if error

geo/geo.go
bufrr: I tested before, cause `range` will passing value not pointer.

yilunzhang: But now all provider is already pointer, so I think it should work. You can test again now

bufrr: In this case i need to change the value of elems inside slice,but elem `p` here is the copy of the value, not a pointer of value.

yilunzhang: `p` is already the pointer when you add it in `AddProvider`, see this example: https://play.golang.org/p/-QlpEpMznc_e

tuna.go
yilunzhang: Did you forget to remove this part?

geo/geo.go
```diff
@@ -0,0 +1,246 @@
...
+		}
+	}
+
+	ip2c := NewIP2CProvider()
+	ip2c.MaybeUpdate()
+	f.Providers = append(f.Providers, ip2c)
+}
+
+func (f *IPFilter) UpdateDataFile() {
+	for {
```
yilunzhang, line 228: This function should exit when tuna stops

geo/geo.go
```diff
@@ -0,0 +1,246 @@
...
+		time.Sleep(10 * time.Second)
+		for idx, p := range f.Providers {
+			if len(p.FileName()) == 0 {
+				continue
+			}
+			if !p.NeedUpdate() {
+				continue
+			}
+
+			err := p.MaybeUpdate()
```
yilunzhang, line 238: The lock should include both `NeedUpdate()` and `MaybeUpdate()`

geo/geo.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+func (f *IPFilter) UpdateDataFile() {
+	for {
+		time.Sleep(10 * time.Second)
+		for idx, p := range f.Providers {
+			if len(p.FileName()) == 0 {
+				continue
+			}
+			if !p.NeedUpdate() {
```
yilunzhang, line 234: Is this needed? `NeedUpdate()` is called inside `MaybeUpdate()` protected by lock, but this call here is not protected by lock

geo/aws.go
```diff
@@ -0,0 +1,172 @@
...
+	AWSFile    = "aws-ip.json"
+)
+
+type AWSProvider struct {
+	Info     *AWSGeoInfo
+	fileName string
+	url      string
+	expire   time.Duration
+	ready    bool
+	mux      sync.Mutex
```
yilunzhang, line 26: This lock cannot solve the problem because different tuna instance will have different lock

geo/geo.go
```diff
@@ -0,0 +1,246 @@
...
+		err = mm.MaybeUpdate()
+		if err != nil {
+			log.Print(err)
+		} else {
+			f.Providers = append(f.Providers, mm)
+		}
+	}
+
+	ip2c := NewIP2CProvider()
+	ip2c.MaybeUpdate()
```
yilunzhang, line 223: Should also handle error here even if currently err is always nil.

entry.go
```diff
@@ -99,6 +99,11 @@ func (te *TunaEntry) Start(shouldReconnect bool) error {
 		log.Printf("Serving %s on localhost udp port %v", te.Service.Name, udpPorts)
 	}
 
+	if !te.IsServer && te.ServiceInfo.IPFilter.NeedGeoInfo() {
+		te.ServiceInfo.IPFilter.AddProvider()
```
yilunzhang, line 103: It's better to use
```go
go func() {
    te.ServiceInfo.IPFilter.AddProvider()
    te.ServiceInfo.IPFilter.UpdateDataFile()
}
```
so that tuna can provide service before db update finish

geo/geo.go
```diff
@@ -0,0 +1,246 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return b, nil
+}
+
+func (f *IPFilter) AddProvider() {
+	if f.DownloadDB {
+		aws := NewAWSProvider(f.DBPath)
+		err := aws.MaybeUpdate()
```
yilunzhang, line 198: Currently `AddProvider()` will take very long to return, and tuna cannot start to provide service (blocked) before it returns. I think it's better not to call `MaybeUpdate()` in `AddProvider()`, and move `time.Sleep` in `UpdateDataFile()` to the end of for loop so update will not block `AddProvider`

geo/geo.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	ip2c := NewIP2CProvider()
+	ip2c.MaybeUpdate()
+	f.Providers = append(f.Providers, ip2c)
+}
+
+func (f *IPFilter) UpdateDataFile() {
+	for {
+		time.Sleep(10 * time.Second)
```
yilunzhang, line 229: 10s might be too short since the db is rarely updated. Something like 1hr might be better.

entry.go
```diff
@@ -36,6 +36,8 @@ type EntryConfiguration struct {
 	ReverseSubscriptionPrefix   string                 `json:"reverseSubscriptionPrefix"`
 	ReverseSubscriptionDuration int32                  `json:"reverseSubscriptionDuration"`
 	ReverseSubscriptionFee      string                 `json:"reverseSubscriptionFee"`
+	DBPath                      string                 `json:"dbpath"`
```
yilunzhang, line 39: Suggest rename to `GeoDBPath`

entry.go
```diff
@@ -36,6 +36,8 @@ type EntryConfiguration struct {
...
+	DownloadDB                  bool                   `json:"downloadDB"`
```
yilunzhang, line 40: Suggest rename to `DownloadGeoDB`

entry.go
```diff
@@ -99,6 +101,12 @@ func (te *TunaEntry) Start(shouldReconnect bool) error {
...
 
+	geoCloseChan := make(chan struct{})
+	if !te.IsServer && te.ServiceInfo.IPFilter.NeedGeoInfo() {
+		te.ServiceInfo.IPFilter.AddProvider(te.config.DownloadDB, te.config.DBPath)
+		go te.ServiceInfo.IPFilter.UpdateDataFile(geoCloseChan)
+	}
+
 	for {
 		if te.IsClosed() {
 			return nil
```
yilunzhang, line 112: need to close geoCloseChan as well here

exit.go
```diff
@@ -10,6 +10,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/nknorg/tuna/geo"
```
yilunzhang, line 13: Did you run `gofmt`?

entry.go
```diff
@@ -99,6 +101,12 @@ func (te *TunaEntry) Start(shouldReconnect bool) error {
 		log.Printf("Serving %s on localhost udp port %v", te.Service.Name, udpPorts)
 	}
 
+	geoCloseChan := make(chan struct{})
+	if !te.IsServer && te.ServiceInfo.IPFilter.NeedGeoInfo() {
+		te.ServiceInfo.IPFilter.AddProvider(te.config.DownloadDB, te.config.DBPath)
+		go te.ServiceInfo.IPFilter.UpdateDataFile(geoCloseChan)
+	}
+
```
yilunzhang, line 109: Better to use `defer close(geoCloseChan)` here

exit.go
```diff
@@ -431,6 +435,12 @@ func (te *TunaExit) Start() error {
 func (te *TunaExit) StartReverse(shouldReconnect bool) error {
 	defer te.Close()
 
+	geoCloseChan := make(chan struct{})
+	if !te.IsServer && te.ServiceInfo.IPFilter.NeedGeoInfo() {
+		te.ServiceInfo.IPFilter.AddProvider(te.config.DownloadDB, te.config.DBPath)
+		go te.ServiceInfo.IPFilter.UpdateDataFile(geoCloseChan)
+	}
+
```
yilunzhang, line 443: Better to use `defer close(geoCloseChan)` here

geo/aws.go
```diff
@@ -0,0 +1,175 @@
...
+func NewAWSProvider(path string) *AWSProvider {
+	p := filepath.Join(path, AWSFile)
+	return &AWSProvider{url: AWSGeoUrl, fileName: p, expire: AWSExpired}
+}
+
+func (p *AWSProvider) MaybeUpdate() error {
+	geoLock.Lock()
+	defer geoLock.Unlock()
+	if !p.NeedUpdate() && p.Info != nil {
+		return errors.New("no need to update aws file")
```
yilunzhang, line 75: Suggest to return nil here because it's not an error. Same for other providers

geo/aws.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	if p.NeedUpdate() {
+		tmpFile, err := ioutil.TempFile("", p.fileName+"-*")
+		if err != nil {
+			return err
+		}
+		err = os.Rename(tmpFile.Name(), p.fileName)
```
yilunzhang, line 82: You are again removing old db before upgrade success. If upgrade fails, old db will not longer be available. Same for other providers.

geo/geo.go
```diff
@@ -0,0 +1,234 @@
...
+		return true
+	}
+	return false
+}
+
+type IPFilter struct {
+	Allow      []Location `json:"allow"`
+	Disallow   []Location `json:"disallow"`
+	providers  []GeoProvider
+	CloseChan  chan struct{}
```
yilunzhang, line 58: Better to be private field

geo/geo.go
```diff
@@ -0,0 +1,234 @@
...
+		loc = &Location{IP: ip}
+	}
+
+	return f.AllowLocation(loc), nil
+}
+
+func (f *IPFilter) GetLocation(ip string) *Location {
+	for _, p := range f.providers {
+		if p.Ready() {
+			loc := getLocationFromProvider(ip, p.(GeoProvider))
```
yilunzhang, line 115: Why type cast?

geo/geo.go
```diff
@@ -0,0 +1,234 @@
...
+			for idx, p := range f.providers {
+				if len(p.FileName()) == 0 {
+					continue
+				}
+				err := p.MaybeUpdate()
+				if err != nil {
+					log.Print(err)
+					continue
+				}
+				f.providers[idx] = p
```
yilunzhang, line 229: Did you forget to remove this?

cmd/entry.go
```diff
@@ -21,7 +22,7 @@ func (e *EntryCommand) Execute(args []string) error {
 		SubscriptionPrefix:        tuna.DefaultSubscriptionPrefix,
 		ReverseSubscriptionPrefix: tuna.DefaultSubscriptionPrefix,
 	}
-	err := tuna.ReadJSON(e.ConfigFile, config)
+	err := geo.ReadJSON(e.ConfigFile, config)
```
yilunzhang, line 25: It is not ideal to put `ReadJSON` into geo package because it's a common utility function. You can create a `util` package to put these common utility functions.

geo/aws.go
```diff
@@ -0,0 +1,173 @@
...
+		return errors.New("no need to update aws file")
+	}
+	if p.NeedUpdate() {
+		tmpFile, err := ioutil.TempFile("", p.fileName+"-*")
+		if err != nil {
+			return err
+		}
+		err = DownloadJsonFile(p.url, tmpFile.Name())
+		if err != nil {
+			return err
```
yilunzhang, line 83: Need to remove temp file before return

yilunzhang (approved):
Looks good to me

PR #52, opened 2020-11-25
fix null pointer panic when err is nil

bufrr:


exit.go
```diff
@@ -260,10 +263,6 @@ func (te *TunaExit) listenTCP(port int) error {
 			if te.IsClosed() {
 				return
 			}
-			if strings.Contains(err.Error(), "use of closed network connection") {
-				te.Close()
-				return
-			}
 			if err != nil {
```
yilunzhang, line 266: You should move it to here

yilunzhang (approved):

PR #53, opened 2020-12-05
add measurement delay

iheron:


tuna.go
```diff
@@ -465,13 +490,80 @@ func (c *Common) CreateServerConn(force bool) error {
 				allSubscribers[i], allSubscribers[j] = allSubscribers[j], allSubscribers[i]
 			})
 
+			// filter
+			var filterSubscribers filterSubscribers
```
yilunzhang, line 494: It's better to change variable name to be different from variable type name `filterSubscribers`

yilunzhang: Also you can use `make(filterSubscribers, 0, len(allSubscribers))` to set a cap in advance to avoid re-alloc memory during append. It might waste some memory but will reduce the number of memory allocations

tuna.go
```diff
@@ -465,13 +490,80 @@ func (c *Common) CreateServerConn(force bool) error {
...
+			}
+			wg.Wait()
+			sort.Sort(SortByDelay{filterSubscribers})
+
+			take := 10
+			if take > len(filterSubscribers) {
+				take = len(filterSubscribers)
+			}
+			topSubscribers := filterSubscribers[:take]
+			log.Println("select top subscribers")
```
yilunzhang, line 556: Forget to remove debug log

tuna.go
```diff
@@ -465,13 +490,80 @@ func (c *Common) CreateServerConn(force bool) error {
...
+			for _, ele := range topSubscribers {
+				log.Println("address: ", ele.metadata.Ip, ", delay: ", ele.delay)
+			}
+			for _, subscriber := range topSubscribers {
```
yilunzhang, line 560: You can remove the price and ip filter in this loop because they have been filtered before

util/measurement.go
```diff
@@ -0,0 +1,19 @@
+package util
+
+import (
+	"context"
+	"net"
+	"time"
+)
+
+func DelayMeasurement(ctx context.Context, network, address string, timeout time.Duration) (error, time.Duration) {
```
yilunzhang, line 9: The convention in Go is to return error as last return value

tuna.go
```diff
@@ -51,9 +54,10 @@ const (
 	MinTrafficCoverage                     = 0.9
 	TrafficDelay                           = 10 * time.Second
 	MaxNanoPayDelay                        = 30 * time.Second
-	getSubscribersBatchSize                = 32
+	getSubscribersBatchSize                = 128
 	DefaultEncryptionAlgo                  = pb.EncryptionAlgo_ENCRYPTION_NONE
 	subscribeDurationRandomFactor          = 0.1
+	measurementJobPoolLength               = 24
```
yilunzhang, line 60: Suggest rename to `measureLatencyConcurrentWorkers`

tuna.go
```diff
@@ -465,13 +490,80 @@ func (c *Common) CreateServerConn(force bool) error {
...
+							default:
+								log.Println(err)
+								break
+							}
+						}
+						subscriber.delay = delay
+					})
+				}(&filterSubscribers[index])
+			}
+			wg.Wait()
```
yilunzhang, line 548: You need to close `measurementDelayJobChan` here to avoid worker goroutine leak

tuna.go
```diff
@@ -465,13 +490,80 @@ func (c *Common) CreateServerConn(force bool) error {
...
+			sort.Sort(SortByDelay{filterSubscribers})
+
+			take := 10
```
yilunzhang, line 551: Why not use all `filterSubscribers`? Since you have filtered by ip and price, it is likely that the first result in `filterSubscribers` will succeed.

util/measurement.go
```diff
@@ -0,0 +1,19 @@
...
+	now := time.Now()
+	d := net.Dialer{Timeout: timeout}
+	_, err := d.DialContext(ctx, network, address)
```
yilunzhang, line 12: Since you don't cancel the context here, I think you don't need the context at all and can simply use `d.Dial()` or even `net.DialTimeout`

tuna.go
iheron: its not debug log

yilunzhang (approved):
Looks good to me

PR #56, opened 2020-12-09
add nkn filter to connect

iheron:


tuna.go
```diff
@@ -82,6 +83,7 @@ type Common struct {
 	ReverseMetadata    *pb.ServiceMetadata
 	OnConnect          *OnConnect
 	IsServer           bool
+	RemoteNknAddress   string
```
yilunzhang, line 86: This property is never used, we can remove it

tuna.go
```diff
@@ -476,29 +491,49 @@ func (c *Common) CreateServerConn(force bool) error {
...
-			if err != nil {
-				return err
-			}
+			if c.ServiceInfo.NknFilter != nil && len(c.ServiceInfo.NknFilter.Allow) > 0 { // nkn client filter
+				nknFilterLength := len(c.ServiceInfo.NknFilter.Allow)
+				subscriberRaw = make(map[string]string, nknFilterLength)
+				allSubscribers = make([]string, 0, nknFilterLength)
+				for _, aFilter := range c.ServiceInfo.NknFilter.Allow {
+					subscription, err := c.Wallet.GetSubscription(topic, aFilter.Address)
+					if err != nil {
```
yilunzhang, line 504: We should print and `continue` here instead of return. If we get error for all addresses in whitelist, then we should return error

yilunzhang (approved):

PR #59, opened 2020-12-17
update ip filter use CIDR

iheron:


geo/geo.go
```diff
@@ -39,7 +41,21 @@ func (l *Location) Empty() bool {
...
-		if location.IP == l.IP {
+		matched, err := regexp.MatchString(`/\d{1,2}`, l.IP)
+		if err != nil {
+			log.Println(err)
+			return false
+		}
+		if !matched {
+			l.IP += "/32"
+		}
+		_, subnet, err := net.ParseCIDR(l.IP)
```
yilunzhang, line 52: I would suggest we add a field `cidr` in `Location` and cache the subnet so we only parse it once

tuna.go
```diff
@@ -55,13 +55,13 @@ const (
...
-	getSubscribersBatchSize                    = 128
+	getSubscribersBatchSize                    = 512
 	DefaultEncryptionAlgo                      = pb.EncryptionAlgo_ENCRYPTION_NONE
 	subscribeDurationRandomFactor              = 0.1
-	measureLatencyConcurrentWorkers            = 24
+	measureLatencyConcurrentWorkers            = 64
 	measureBandwidthConcurrentWorkers          = 8
 	measureDelayTopDelayCount                  = 32
-	measurementBytesDownlink                   = 128 << 10
+	measurementBytesDownLink                   = 128 << 11
```
yilunzhang, line 64: It's better to use `256 << 10` because `<< 10` means `KB`

yilunzhang (approved):

PR #61, opened 2021-01-11
add measure cache

iheron:


storage/storage.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func (s *Storage) Delete(key string) {
+	s.Lock()
+	defer s.Unlock()
+	delete(s.data, key)
+}
+
+func (s *Storage) Len() int {
+	return len(s.data)
```
yilunzhang, line 42: Need rlock here

storage/measure.go
```diff
@@ -0,0 +1,221 @@
...
+	if !isExists {
+		err := util.WriteJSON(s.favoriteFilePath, map[string]*FavoriteNode{})
+		if err != nil {
+			return err
+		}
+	}
+	var favoriteData map[string]*FavoriteNode
+	err := util.ReadJSON(s.favoriteFilePath, &favoriteData)
+	if err != nil {
+		return err
```
yilunzhang, line 76: Error here means json is invalid. Should we remove the file and re-create? Same for bad node

storage/measure.go
```diff
@@ -0,0 +1,221 @@
...
+		val.ExpiresAt = time.Now().Add(badExpired).Unix()
+	}
+
+	if val.MaskSize == 0 {
+		val.MaskSize = maskSize
+	}
+
+	_, subnet, err := net.ParseCIDR(fmt.Sprintf("%s/%d", key, val.MaskSize))
+	if err != nil {
+		log.Println(err)
```
yilunzhang, line 201: Should we return here? If `err != nil`, `subnet` is probably `nil` and the next part will panic

storage/measure.go
```diff
@@ -0,0 +1,221 @@
...
+		s.BadNodes[subnet.String()][key] = val
+	} else {
+		s.BadNodes[subnet.String()] = map[string]*BadNode{
+			val.IP: val,
+		}
+	}
+}
+
+func (s *MeasureStorage) GetBadNodes() []string {
+	var results []string
```
yilunzhang, line 214: Need lock here

storage/measure.go
```diff
@@ -0,0 +1,221 @@
...
+	if _, ok := s.BadNodes[subnet.String()]; ok {
+		s.BadNodes[subnet.String()][key] = val
+	} else {
+		s.BadNodes[subnet.String()] = map[string]*BadNode{
+			val.IP: val,
+		}
+	}
+}
+
+func (s *MeasureStorage) GetBadNodes() []string {
```
yilunzhang, line 213: Suggest rename to `GetBadCIDR`

storage/measure.go
```diff
@@ -0,0 +1,221 @@
...
+	var results []string
+	for k, v := range s.BadNodes {
+		if len(v) > 3 {
```
yilunzhang, line 216: It's better to move 3 as a constant like `BadCIDRMinIP`

tests/storage_test.go
```diff
@@ -0,0 +1,45 @@
...
+		log.Println(err)
+		return
+	}
+
+	err = measureStorage.SaveFavoriteNodes()
+	if err != nil {
+		log.Println(err)
+	}
+	//
+	//measureStorage.BadNodes = append(measureStorage.BadNodes, &storage.BadNode{
```
yilunzhang, line 35: Should remove unused test code

tuna.go
```diff
@@ -684,21 +700,37 @@ func (c *Common) filterSubscribers(allSubscribers []string, subscriberRaw map[st
 			continue
 		}
 
+		if c.measureStorage != nil { // disallow bad nodes
+			nodes := c.measureStorage.GetBadNodes()
```
yilunzhang, line 704: This is called for every subscriber. It's better to call it before the for loop

tuna.go
```diff
@@ -777,21 +809,49 @@ func (c *Common) measureBandwidth(delayMeasuredSubs types.Nodes, n int) types.No
...
+					}
+					c.measureStorage.AddFavoriteNode(sub.Metadata.Ip, &storage.FavoriteNode{
+						IP:           sub.Metadata.Ip,
+						Address:      sub.Address,
+						Metadata:     base64.StdEncoding.EncodeToString(metadata),
+						Delay:        sub.Delay,
+						MinBandwidth: min / 1000,
+						MaxBandwidth: max / 1000,
+					})
+					err = c.measureStorage.SaveFavoriteNodes()
```
yilunzhang, line 841: I think better to let `AddFavoriteNode` return a bool indicating whether node is added or not, and only save if node is added

tuna.go
```diff
@@ -777,21 +809,49 @@ func (c *Common) measureBandwidth(delayMeasuredSubs types.Nodes, n int) types.No
...
+					metadata, err := proto.Marshal(sub.Metadata)
+					if err != nil {
+						log.Println(err)
+					}
+					c.measureStorage.AddFavoriteNode(sub.Metadata.Ip, &storage.FavoriteNode{
+						IP:           sub.Metadata.Ip,
+						Address:      sub.Address,
+						Metadata:     base64.StdEncoding.EncodeToString(metadata),
+						Delay:        sub.Delay,
+						MinBandwidth: min / 1000,
```
yilunzhang, line 838: It's better to use 1024

tuna.go
```diff
@@ -777,21 +809,49 @@ func (c *Common) measureBandwidth(delayMeasuredSubs types.Nodes, n int) types.No
...
 					}
 					return
 				}
 
 				log.Printf("address: %s, bandwidth: %f - %f KB/s, time: %s", addr, min/1000, max/1000, dur)
 
+				if c.measureStorage != nil {
+					metadata, err := proto.Marshal(sub.Metadata)
+					if err != nil {
+						log.Println(err)
```
yilunzhang, line 831: Should not use metadata if err != nil

tuna.go
```diff
@@ -656,6 +690,12 @@ func (c *Common) filterSubscribers(allSubscribers []string, subscriberRaw map[st
 		log.Fatalf("Parse price of service error: %v", err)
 	}
 	filterSubs := make(types.Nodes, 0, len(allSubscribers))
+
+	var nodes []string
+	if c.measureStorage != nil {
+		nodes = c.measureStorage.GetAvoidCIDR()
```
yilunzhang, line 696: It's more efficient to parse cidr here and get `[]*net.IPNet`

tuna.go
```diff
@@ -777,21 +832,55 @@ func (c *Common) measureBandwidth(delayMeasuredSubs types.Nodes, n int) types.No
...
 				}
 
 				log.Printf("address: %s, bandwidth: %f - %f KB/s, time: %s", addr, min/1000, max/1000, dur)
 
+				if c.measureStorage != nil {
+					var metadataString string
+					metadata, err := proto.Marshal(sub.Metadata)
+					if err != nil {
+						log.Println(err)
+						metadataString = ""
```
yilunzhang, line 856: You should not add favorite node if err != nil because the current logic assumes metadata is not empty

PR #104, opened 2021-07-13
Test conn on reverse exit before choose entry node

yilunzhang:
Signed-off-by: Yilun <zyl.skysniper@gmail.com>

### DIFF
--- a/exit.go
+++ b/exit.go
@@ -549,6 +549,25 @@ func (te *TunaExit) StartReverse(shouldReconnect bool) error {
 			continue
 		}
 
+		reverseIP := tcpConn.RemoteAddr().(*net.TCPAddr).IP
+		reverseTCP := reverseMetadata.ServiceTcp
+		if len(reverseTCP) > 0 {
+			go func() {
+				net.DialTimeout(tcp, fmt.Sprintf("%s:%d", reverseIP.String(), reverseTCP[0]), defaultReverseTestTimeout)
+			}()
+
+			session.SetDeadline(time.Now().Add(defaultReverseTestTimeout))
+
+			_, err = session.AcceptStream()
+			if err != nil {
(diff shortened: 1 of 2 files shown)


exit.go
```diff
@@ -549,6 +549,25 @@ func (te *TunaExit) StartReverse(shouldReconnect bool) error {
...
 		}
 
+		reverseIP := tcpConn.RemoteAddr().(*net.TCPAddr).IP
+		reverseTCP := reverseMetadata.ServiceTcp
+		if len(reverseTCP) > 0 {
+			go func() {
+				net.DialTimeout(tcp, fmt.Sprintf("%s:%d", reverseIP.String(), reverseTCP[0]), defaultReverseTestTimeout)
+			}()
+
+			session.SetDeadline(time.Now().Add(defaultReverseTestTimeout))
```
bufrr, line 559: Could we call `SetDeadline` when session init?

exit.go
```diff
@@ -549,6 +549,25 @@ func (te *TunaExit) StartReverse(shouldReconnect bool) error {
...
+
+			_, err = session.AcceptStream()
+			if err != nil {
+				log.Println("Couldn't accept stream test conn from reverse entry:", err)
+				time.Sleep(1 * time.Second)
+				continue
+			}
+
+			session.SetDeadline(time.Time{})
```
bufrr, line 568: Is this required to set it back?

exit.go
yilunzhang: This is to prevent `session.Accept()` not returning (if entry port is not open)

exit.go
yilunzhang: Yes, otherwise `session.Accept()` will always return error after 3s

bufrr (approved):

PR #117, opened 2022-03-21
Better handling of txn fee and subscription failure

yilunzhang:
* Check sub status periodically to handle sub failure
* Use sub fee only if wallet has enough balance
* Add config to replace txpool when subscribing

### DIFF
--- a/tuna.go
+++ b/tuna.go
@@ -1270,7 +1270,23 @@ func UpdateMetadata(
 				return
 			}
 
-			addToSubscribeQueue(client, identifier, topic, int(subscriptionDuration), string(metadataRaw), &nkn.TransactionConfig{Fee: subscriptionFee})
+			subFee, err := common.StringToFixed64(subscriptionFee)
+			if err != nil {
+				log.Println("Parse subscription fee error:", err)
+			}
+
+			if subFee > 0 {
+				balance, err := client.Balance()
+				if err != nil {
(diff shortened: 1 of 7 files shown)


tuna.go
```diff
@@ -1225,50 +1226,77 @@ func UpdateMetadata(
...
+			addToSubscribeQueue(client, identifier, topic, int(subscriptionDuration), string(metadataRaw), &nkn.TransactionConfig{Fee: subFee.String()}, subscriptionReplaceTxPool)
 
-			maxSubDuration := float64(sub.ExpiresAt-height) * float64(config.ConsensusDuration)
-			nextSub = time.After(time.Duration((1 - rand.Float64()*subscribeDurationRandomFactor) * maxSubDuration))
-		}()
+			nextSub = time.After(time.Duration((1 - rand.Float64()*subscribeDurationRandomFactor) * float64(subInterval)))
 
-		for {
 			select {
 			case <-nextSub:
```
bufrr, line 1295: Do we need an other `<-nextSub` here?

bufrr, line 1295: Can we keep only `maxCheckSubscribeInterval` as check interval?

yilunzhang: We can't because `subInterval` is configurable and can be less than `maxCheckSubscribeInterval`

bufrr (approved):

PR #133, opened 2023-03-07
mv udp code to the root

bufrr:
mv udp code to the root dir of tuna

### DIFF
--- a/cmd/entry.go
+++ b/cmd/entry.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"github.com/nknorg/tuna/udp"
 	"log"
 	"strings"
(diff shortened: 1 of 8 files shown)


yilunzhang (commented):
I think it's better to rename type Conn to UDPConn

bufrr:
> I think it's better to rename type Conn to UDPConn

Using the same name with `net.UDPConn` may cause  confusion for user

yilunzhang:
It's fine because one is net.UDPConn and one is tuna.UDPConn. There is also a net.Conn which has the same name

yilunzhang (approved):

PR #135, opened 2023-03-10
fix udp write to wrong addr when client under NAT

bufrr:


entry.go
```diff
@@ -767,13 +765,23 @@ func StartReverse(config *EntryConfiguration, wallet *nkn.Wallet) error {
 
 					te.SetServerTCPConn(encryptedConn)
 
-					if metadata.UdpPort > 0 {
-						ip, _, err := net.SplitHostPort(encryptedConn.RemoteAddr().String())
+					if len(metadata.ServiceUdp) > 0 {
+						var clientAddr string
+						addrToKeyLock.RLock()
+						for u, v := range addrToKey {
```
yilunzhang, line 771: It's better to use another `keyToAddr` map for O(1) look up time

yilunzhang (approved):

PR #147, opened 2023-04-04
Check wallet balance in CreateServerConn

billfort:
1. Add minBalance in entry and exit config;
2. Check wallet balance in tuna.CreateServerConn. If wallet balance is less than config minBalance, connection would be rejected.

### DIFF
--- a/config.go
+++ b/config.go
@@ -35,6 +35,7 @@ const (
 	maxMeasureBandwidthTimeout               = 30 * time.Second
 	nanoPayClaimerLinger                     = 24 * time.Hour
 	maxCheckSubscribeInterval                = time.Hour
+	defaultMinBalance                        = "0.0" // default minimum wallet balance for use tuna service
 )
 
 type EntryConfiguration struct {
@@ -72,6 +73,7 @@ type EntryConfiguration struct {
 	TcpDialContext                   func(ctx context.Context, network, addr string) (net.Conn, error) `json:"-"`
 	HttpDialContext                  func(ctx context.Context, network, addr string) (net.Conn, error) `json:"-"`
 	WsDialContext                    func(ctx context.Context, network, addr string) (net.Conn, error) `json:"-"`
+	MinBalance                       string                                                            `json:"minBalance"`
 }
 
 var defaultEntryConfiguration = EntryConfiguration{
@@ -85,6 +87,7 @@ var defaultEntryConfiguration = EntryConfiguration{
 	ReverseServiceName:             DefaultReverseServiceName,
 	ReverseMinFlushAmount:          defaultNanoPayMinFlushAmount,
 	ReverseServiceListenIP:         defaultReverseServiceListenIP,
+	MinBalance:                     defaultMinBalance,
 }
 
 func DefaultEntryConfig() *EntryConfiguration {
@@ -130,6 +133,7 @@ type ExitConfiguration struct {
 	TcpDialContext                 func(ctx context.Context, network, addr string) (net.Conn, error) `json:"-"`
 	HttpDialContext                func(ctx context.Context, network, addr string) (net.Conn, error) `json:"-"`
 	WsDialContext                  func(ctx context.Context, network, addr string) (net.Conn, error) `json:"-"`
+	ReverseMinBalance              string                                                            `json:"reverseMinBalance"`
 }
 
 var defaultExitConfiguration = ExitConfiguration{
@@ -142,6 +146,7 @@ var defaultExitConfiguration = ExitConfiguration{
 	MinFlushAmount:                 defaultNanoPayMinFlushAmount,
 	ReverseSubscriptionPrefix:      DefaultSubscriptionPrefix,
 	ReverseServiceName:             DefaultReverseServiceName,
+	ReverseMinBalance:              defaultMinBalance,
 }
 
 func DefaultExitConfig() *ExitConfiguration {
--- a/tuna.go
+++ b/tuna.go
@@ -146,6 +146,8 @@ type Common struct {
 
 	reverseBytesExitToEntry map[string][]uint64
 	reverseBytesEntryToExit map[string][]uint64
+
+	minBalance common.Fixed64 // minimum wallet balance requirement for connecting
 }
 
 func NewCommon(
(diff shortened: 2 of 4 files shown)


config.go
```diff
@@ -130,6 +133,7 @@ type ExitConfiguration struct {
 	TcpDialContext                 func(ctx context.Context, network, addr string) (net.Conn, error) `json:"-"`
 	HttpDialContext                func(ctx context.Context, network, addr string) (net.Conn, error) `json:"-"`
 	WsDialContext                  func(ctx context.Context, network, addr string) (net.Conn, error) `json:"-"`
+	MinBalance                     float64                                                           `json:"minBalance"`
```
yilunzhang, line 136: Should be `ReverseMinBalance` because it's used in reverse mode

tuna.go
```diff
@@ -745,6 +750,21 @@ func (c *Common) CreateServerConn(force bool) error {
 				return err
 			}
 
+			entryToExitMaxPrice, exitToEntryMaxPrice, err := ParsePrice(c.ServiceInfo.MaxPrice)
+			if err != nil {
+				return err
+			}
+			if entryToExitMaxPrice > 0 || exitToEntryMaxPrice > 0 {
+				balance, err := c.Wallet.Balance()
```
yilunzhang, line 758: Better to use `c.Client.BalanceByAddress(c.Wallet.Address())` because it's more reliable

config.go
```diff
@@ -72,6 +73,7 @@ type EntryConfiguration struct {
 	TcpDialContext                   func(ctx context.Context, network, addr string) (net.Conn, error) `json:"-"`
 	HttpDialContext                  func(ctx context.Context, network, addr string) (net.Conn, error) `json:"-"`
 	WsDialContext                    func(ctx context.Context, network, addr string) (net.Conn, error) `json:"-"`
+	MinBalance                       float64                                                           `json:"minBalance"`
```
yilunzhang, line 76: Better to use string type

tuna.go
```diff
@@ -146,6 +146,8 @@ type Common struct {
 
 	reverseBytesExitToEntry map[string][]uint64
 	reverseBytesEntryToExit map[string][]uint64
+
+	minBalance float64 // minimum wallet balance requirement for connecting
```
yilunzhang, line 150: Use `Fixed64` type and parse it with `StringToFixed64` method

tuna.go
```diff
@@ -745,6 +750,21 @@ func (c *Common) CreateServerConn(force bool) error {
...
+			if err != nil {
+				return err
+			}
+			if entryToExitMaxPrice > 0 || exitToEntryMaxPrice > 0 {
+				balance, err := c.Wallet.Balance()
+				if err != nil {
+					return err
+				}
+				fBalance, _ := strconv.ParseFloat(balance.Fixed64.String(), 64)
+				if fBalance <= c.minBalance {
```
yilunzhang, line 763: Should be `<`

tuna.go
```diff
@@ -745,6 +750,21 @@ func (c *Common) CreateServerConn(force bool) error {
 				return err
 			}
 
+			entryToExitMaxPrice, exitToEntryMaxPrice, err := ParsePrice(c.ServiceInfo.MaxPrice)
```
yilunzhang, line 753: Only need to check these if `minBalance > 0`

tuna.go
```diff
@@ -745,6 +750,21 @@ func (c *Common) CreateServerConn(force bool) error {
...
+			if err != nil {
+				return err
+			}
+			if entryToExitMaxPrice > 0 || exitToEntryMaxPrice > 0 {
+				balance, err := c.Wallet.Balance()
+				if err != nil {
+					return err
```
yilunzhang, line 760: I think we shouldn't return when failed to get balance so tuna won't be affected by on-chain incidents.

config.go
billfort: Yes, modified

tuna.go
billfort: Yes, modified.

config.go
billfort: Yes, modified.

tuna.go
billfort: Yes, modified.

tuna.go
billfort: Yes, modified.

tuna.go
billfort: Yes, modified.

tuna.go
billfort: Yes, modified.

tuna.go
```diff
@@ -250,6 +253,7 @@ func NewCommon(
 		udpWriteChan:  make(chan []byte),
 		connReadyChan: make(map[string]chan struct{}),
 	}
+	c.minBalance, _ = common.StringToFixed64(minBalance)
```
yilunzhang, line 256: Always handle error

tuna.go
```diff
@@ -745,6 +749,19 @@ func (c *Common) CreateServerConn(force bool) error {
...
 			}
 
+			if c.minBalance > 0 {
+				entryToExitMaxPrice, exitToEntryMaxPrice, err := ParsePrice(c.ServiceInfo.MaxPrice)
+				if err != nil {
+					return err
+				}
+				if entryToExitMaxPrice > 0 || exitToEntryMaxPrice > 0 {
+					balance, err := c.Client.BalanceByAddress(c.Wallet.Address())
+					if err == nil && balance.Fixed64 < c.minBalance {
```
yilunzhang, line 759: Better use `ToFixed64()`

tuna.go
billfort: Yes, modified.

tuna.go
billfort: Yes, modified.

tuna.go
```diff
@@ -745,6 +752,19 @@ func (c *Common) CreateServerConn(force bool) error {
...
 			}
 
+			if c.minBalance > 0 {
+				entryToExitMaxPrice, exitToEntryMaxPrice, err := ParsePrice(c.ServiceInfo.MaxPrice)
+				if err != nil {
+					return err
+				}
+				if entryToExitMaxPrice > 0 || exitToEntryMaxPrice > 0 {
+					balance, err := c.Client.BalanceByAddress(c.Wallet.Address())
+					if err == nil && balance.ToFixed64() < c.minBalance {
```
yilunzhang, line 762: I would suggest we log the error if not nil

billfort: Yes, modified.

yilunzhang (approved):

PR #148, opened 2023-06-26
Tcp deadlock fix

bufrr:
Fix deadlock when udp connection block tcp, and using `SetLinger(0)` to avoid to many CLOSE_WAIT / TIME_WAIT issue

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -317,7 +317,9 @@ func (te *TunaEntry) Close() {
 	for _, conn := range te.serviceConn {
 		Close(conn)
 	}
-	te.session.Close()
+	if te.session != nil {
+		te.session.Close()
+	}
 	te.OnConnect.close()
 }
 
@@ -426,6 +428,13 @@ func (te *TunaEntry) listenTCP(ip net.IP, ports []uint32) ([]uint32, error) {
 		go func() {
 			for {
 				conn, err := listener.Accept()
+				if c, ok := conn.(*net.TCPConn); ok {
+					err := c.SetLinger(0)
+					if err != nil {
+						log.Println("Couldn't set linger:", err)
+						continue
+					}
+				}
 				if err != nil {
 					if te.IsClosed() {
 						return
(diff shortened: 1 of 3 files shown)


entry.go
```diff
@@ -426,6 +428,13 @@ func (te *TunaEntry) listenTCP(ip net.IP, ports []uint32) ([]uint32, error) {
 		go func() {
 			for {
 				conn, err := listener.Accept()
+				if c, ok := conn.(*net.TCPConn); ok {
+					err := c.SetLinger(0)
+					if err != nil {
+						log.Println("Couldn't set linger:", err)
+						return
```
yilunzhang, line 435: Why `return` here but `continue` in reverse (L704)?

entry.go
```diff
@@ -657,17 +649,11 @@ func StartReverse(config *EntryConfiguration, wallet *nkn.Wallet) error {
...
-				keyToAddr[k] = from.String()
-				keyToAddrLock.Unlock()
+				addrToKey.Store(from.String(), k)
+				keyToAddr.Store(k, from.String())
 
-				udpReadyLock.RLock()
-				closeChan(udpReady[k])
-				udpReadyLock.RUnlock()
+				c, _ := udpReady.Load(k)
+				closeChan(c.(chan struct{}))
```
yilunzhang, line 656: should only do type conversion if load success

entry.go
```diff
@@ -795,17 +781,15 @@ func StartReverse(config *EntryConfiguration, wallet *nkn.Wallet) error {
...
 
 					if len(metadata.ServiceUdp) > 0 {
-						tcpReadyLock.RLock()
-						<-tcpReady[connKey]
-						tcpReadyLock.RUnlock()
+						tr, _ := tcpReady.Load(connKey)
+						<-tr.(chan struct{})
+
+						ur, _ := udpReady.Load(connKey)
+						<-ur.(chan struct{})
```
yilunzhang, line 788: should only do type conversion if load success

entry.go
```diff
@@ -795,17 +781,15 @@ func StartReverse(config *EntryConfiguration, wallet *nkn.Wallet) error {
...
 
-						udpReadyLock.RLock()
-						<-udpReady[connKey]
-						udpReadyLock.RUnlock()
+						addr, _ := keyToAddr.Load(connKey)
+						clientAddr := addr.(string)
```
yilunzhang, line 791: should only do type conversion if load success

entry.go
```diff
@@ -795,28 +787,32 @@ func StartReverse(config *EntryConfiguration, wallet *nkn.Wallet) error {
...
-						go func() {
+							addr, ok := keyToAddr.Load(connKey)
+							if !ok {
+								return
+							}
+							clientAddr := addr.(string)
+
+							ip, portStr, err := net.SplitHostPort(clientAddr)
+							if err != nil {
+								log.Printf("parse host error: %v\n", err)
```
yilunzhang, line 811: return?

entry.go
```diff
@@ -828,22 +824,17 @@ func StartReverse(config *EntryConfiguration, wallet *nkn.Wallet) error {
...
 									}
-									atomic.AddUint64(&te.Common.reverseBytesExitToEntry[key][data[2]], uint64(n))
+									atomic.AddUint64(&te.Common.reverseBytesExitToEntry[key.(string)][data[2]], uint64(n))
 								case <-te.udpCloseChan:
 									return
 								}
 							}
 						}()
-
-						te.SetServerUDPReadChan(udpReadChan)
```
yilunzhang, line 845: Why remove these?

bufrr: > Why remove these?

Both init within `NewCommon` already

yilunzhang (approved):
LGTM

PR #151, opened 2023-07-18
sending heart beat msg from client to server periodically

bufrr:
1. Send UDP heartbeat packet from client to server 
2. Optimize lock type


util.go
```diff
@@ -191,6 +194,25 @@ func writeUDPConnMetadata(conn UDPConn, addr *net.UDPAddr, connMetadata *pb.Conn
 	return err
 }
 
+func sendPingMsg(conn UDPConn, num *uint64) {
+	pingMsg := new(pb.ConnectionMetadata)
+	pingMsg.IsPing = true
+	b, _ := proto.Marshal(pingMsg)
```
yilunzhang, line 200: What about we make `writeUDPConnMetadata` return n?

util.go
```diff
@@ -191,6 +194,25 @@ func writeUDPConnMetadata(conn UDPConn, addr *net.UDPAddr, connMetadata *pb.Conn
...
+
+	for {
+		err := writeUDPConnMetadata(conn, nil, pingMsg)
```
yilunzhang, line 203: Will this function call return error when conn is closed? We need to make sure this goroutine doesn't leak. Need to test this.

util.go
```diff
@@ -191,6 +194,25 @@ func writeUDPConnMetadata(conn UDPConn, addr *net.UDPAddr, connMetadata *pb.Conn
...
+		if err != nil {
+			log.Println("write udp ping msg error:", err)
+			break
+		}
+		if num != nil {
+			atomic.AddUint64(num, uint64(len(b)))
```
yilunzhang, line 209: Why add to bytesWrite? It's not counted in bytesRead

entry.go
```diff
@@ -617,14 +618,14 @@ func StartReverse(config *EntryConfiguration, wallet *nkn.Wallet) error {
 				continue
 			}
 			if bytes.Equal(buffer[:PrefixLen], []byte{PrefixLen - 1: 0}) && n > PrefixLen {
-				if encrypted {
-					continue
-				}
 				connMetadata, err := parseUDPConnMetadata(buffer[PrefixLen:n])
```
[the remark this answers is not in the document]
bufrr: Yes, the ping msg is encrypted since it will be sent after the UDP conn encrypted

> metadata will be parsed even if `encrypted == true`, is that expected?

yilunzhang (approved):

PR #153, opened 2023-08-02
udp: fix read from closed udp conn

bufrr:


tuna.go
```diff
@@ -415,8 +415,9 @@ func (c *Common) startUDPReaderWriter(conn *EncryptUDPConn, toAddr *net.UDPAddr,
 			n, from, encrypted, err = conn.ReadFromUDPEncrypted(buffer)
 			if err != nil {
 				log.Println("Couldn't receive data:", err)
-				if strings.Contains(err.Error(), "use of closed network connection") {
-					c.udpCloseChan <- struct{}{}
+				if !strings.Contains(err.Error(), "decrypt failed") {
```
yilunzhang, line 418: What about `use of closed network connection` error?

tuna.go
```diff
@@ -415,8 +415,9 @@ func (c *Common) startUDPReaderWriter(conn *EncryptUDPConn, toAddr *net.UDPAddr,
 			n, from, encrypted, err = conn.ReadFromUDPEncrypted(buffer)
 			if err != nil {
 				log.Println("Couldn't receive data:", err)
-				if strings.Contains(err.Error(), "use of closed network connection") {
-					c.udpCloseChan <- struct{}{}
+				if errors.Is(err, io.ErrClosedPipe) {
```
yilunzhang, line 418: `io.ErrClosedPipe` is `io: read/write on closed pipe`, different from `use of closed network connection`

tuna.go
```diff
@@ -1067,7 +1071,8 @@ func measureDelay(ctx context.Context, nodes types.Nodes, concurrentWorkers, num
 				addr := node.Metadata.Ip + ":" + strconv.Itoa(int(node.Metadata.TcpPort))
 				delay, err := tunaUtil.DelayMeasurementContext(ctx, tcp4, addr, timeout, dialContext)
 				if err != nil {
-					if _, ok := err.(net.Error); !ok {
+					var e net.Error
+					if !errors.As(err, &e) {
```
yilunzhang, line 1075: Should this be `errors.Is`?

tuna.go
```diff
@@ -1126,7 +1131,8 @@ func (c *Common) measureBandwidth(ctx context.Context, nodes types.Nodes, n int,
 			}
 			conn, err := dialContext(ctx, tcp4, addr)
 			if err != nil {
-				if _, ok := err.(net.Error); !ok {
+				var e net.Error
+				if !errors.As(err, &e) {
```
yilunzhang, line 1135: Should this be `errors.Is`?

tuna.go
bufrr: > ErrClosedPipe

In this case, `io.ErrClosedPipe` is the error when read from a closed UDP connection

tuna.go
bufrr: > Should this be `errors.Is`?

`erross.As`  try to find if `e` is some kind of `net.Error`

